PR #17, opened 2015-07-14
Fixes mpeg4 avc pts and display aspect ratio when playing software decoded video

reego-fr:


reego-fr:
Hi,
first of all, thank you very much for this project ! I was looking for a software decoding alternative to support interlaced h264 decoding on Windows Phone, you made it easy enough for me to give it a try. 

I got mixed results on my Lumia 1020, most probably because the operation is highly cpu intensive. 

These two commits fixed some of the issues I encountered while testing my sample streams and other mixed video files. I know the changes are small but I hope it can help !

khouzam:
Hi,

Thank you very much for your contributions, we'll take a look at the changes, they seem good, but would you have a couple of sample files that we can test to validate the issues?

reego-fr:
Here are two files that don't play well because of the pts issue : http://1drv.ms/1fmUZxd 
The mkv file contains a mpeg-4 avc encoded video (check force video decode) and the vob file contains mpeg-2 video. I previously though mpeg-2 decoding was fine but it is worth mentioning the issue may appear on other codecs as well. 
Finally, the vob file displays as 4:3 but its correct aspect ratio is 16:9. Media Foundation detects it correctly when decoding h264 by itself (I tested it but don't have samples to share, I don't know about other codecs), so I think it might be needed when doing software video decoding only.

timotiusmargo:
Thanks @reego-fr! These are good stuff :+1:

timotiusmargo:
Hi @reego-fr, would you mind verifying the pull request below?
https://github.com/Microsoft/FFmpegInterop/pull/32

And thanks for the all the good work. We'll review the ts files support soon.

reego-fr:
Hi, as said in the other thread, the pull request looks good to me. 

For reference, and correct my first message, I found out interlaced h264 hardware decoding was working on Windows Phone (on my streams at least). I was probably not using it correctly in my older application !

PR #32, opened 2015-09-23
Fix timestamp correction

timotiusmargo:
When decoding DivX video (and VC1 etc), m_pAvFrame->pkt_pts contains AV_NOPTS_VALUE right after complete frame decoding. The reason why it still works at the moment is the extra timestamp correction logic in MediaSampleProvider that simply use the dts,

av_frame_get_best_effort_timestamp() seems to correctly return the right pts in my tests. With this correction, I no longer have to correct the pts in MediaSampleProvider as well. Could you please verify with your media files and let us know if you have any issue?


reego-fr:
Hi, I will take a look at it.

reego-fr:
Hi,
it looks OK for me and using av_frame_get_best_effort_timestamp seems to be the best way of dealing with this issue as you noted.

Issue #24, opened 2015-08-24
Issue with ts files and two uncompressed audio issues solved 


reego-fr:
Hi,

I encountered some issues while testing various files and hopefully I was able to figure out what was wrong. One of my fixes actually changes the signature of a method. That's why I don't submit a pull request right away, there are many ways this issue can be solved but I took the easy one. 

Sample files can be found here: http://1drv.ms/1fmUZxd
1-Files having an audio stream whose bit depth is not 16 bits won't play correctly. The wav file I provide is 8 bits. The fix is provided by this commit: https://github.com/reego-fr/FFmpegInterop/commit/3d0a3d5a605aa5a6cde093602d4fcd706b57283e
It is worth mentioning that my desktop is actually able to play audio at 32 bits per samples while my phone is limited to 16 bits only. I don't know if it's possible to retrieve this capability at runtime to deliver the best quality and give the phone hardware/software a chance to support 32 bits in the future but for now it does the job. 

2-Some audio streams may not render correctly when decoded by ffmpeg depending on the number of samples per frame. I encountered the issue on dvd_pcm streams. Two of the files I provide exhibit this issue on their pcm stream : the dts sparks vob and vts_04_04 vob. I fixed it in this commit: https://github.com/reego-fr/FFmpegInterop/commit/fb41be9ddab80d4e55da370ff7d177e04ca04431

3-Some files may start at a timestamp greater than 0. Media Foundation most probably expects samples to be delivered within the start (0) and media duration range. Even though it shows as quite tolerant regarding sample timestamps, the issue becomes critical when it comes to seeking in the media's range. 
The sample files to reproduce the issue are the vts_04_04 vob file I provide, and any other vts_xx_02+.vob, as well as ts files but I can't share one of these yet. 
The fix I chose is it this commit: https://github.com/reego-fr/FFmpegInterop/commit/a21b3f231153761d74299efddb72145daa694f98

Please let me know what you think about it and if you want me to submit pull requests. 

Kind regards

reego-fr:
Hi,

I just published three branches each of them holding the fixes for these issues. 

I submitted a pull request for the most mature of them (the 2nd one on pcm_dvd streams). 

I think my fix for the media start time issue is good but I encountered negative pts values (-4) right at the beginning of some files. The maths looks ok so there must be rounding issues. Media Foundation accepts it and plays the file fine so I don't know if I should care about it. 

[...]

PR #35, opened 2015-09-29
Fixing wrong pcm frame length to match the actual number of samples

reego-fr:
Hi, 

as I said in this thread https://github.com/Microsoft/FFmpegInterop/issues/24, I encountered an issue while playing dvd_pcm streams. Two of the files I provide in this folder http://1drv.ms/1fmUZxd exhibit this issue on their pcm stream (the default one) : the dts sparks vob and vts_04_04 vob. 

It looks like av_samples_alloc return value can not reliably be used to get the resampled data range. Instead, my change relies on swr_convert return value to get it.


reego-fr:
Hi,

Sure, I can commit the changes at the end of the day.

I'll share my other findings with you too. It deals with aac adts compressed samples playback. 

Kind regards.

reego-fr:
Hi,

I submitted the changes. The code might need some error checking but it worked ok for me so far...

PR #41, opened 2015-10-13
Aac ADTS compressed samples fix

reego-fr:
Hi,

I encountered issues when playing aac adts compressed samples. Samples having this format can be found in mpeg transport streams (ex: h264+aac) or Shoutcast streams (aac). 
For your convenience, here is a Shoutcast streaming aac found on shoutcast.com : http://5.39.82.140:8144

### DIFF
--- a/FFmpegInterop/Source/FFmpegInteropMSS.cpp
+++ b/FFmpegInterop/Source/FFmpegInteropMSS.cpp
@@ -394,7 +394,14 @@ HRESULT FFmpegInteropMSS::CreateAudioStreamDescriptor(bool forceAudioDecode)
 {
 	if (avAudioCodecCtx->codec_id == AV_CODEC_ID_AAC && !forceAudioDecode)
 	{
-		audioStreamDescriptor = ref new AudioStreamDescriptor(AudioEncodingProperties::CreateAac(avAudioCodecCtx->sample_rate, avAudioCodecCtx->channels, avAudioCodecCtx->bit_rate));
+		if (avAudioCodecCtx->extradata_size == 0)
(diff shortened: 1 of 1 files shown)


reego-fr:
As a side note, 
thanks to your latest changes this is the only change I need to make to the FFmpegInterop project to make it play my streams over rtsp (Media Foundation decoding only). Unlike stream dumps, realtime streaming does not require the samples to be delivered from 0.

timotiusmargo:
Thanks for your contribution @reego-fr! This is good stuff! We're looking forward to see more of your great work.

As a side note, we also recently added unit test that we can use to prevent regression errors. I think it is worthwhile to add a unit test under ```TestCreateFFmpegInteropMSSFromUri.cs``` to get the coverage of the AAC ADTS path.

reego-fr:
Thank you for the encouragements @timotiusmargo. I might have some small changes improving the compatibility a bit that I will be happy to share.  

At the same time, I am having issues enabling background playback for my streams on windows 8.1. I was not able to make it work in my older application either so I wonder if you have any tips or know about issues with the MediaStreamSource. 

I had a look at the tests. I first thought we would only need to check the MediaStreamSource properties for AudioEncodingProperties or AudioStreamDescriptor objects but it apparently does not expose any. I think we can get them from somewhere else but is it what you had in mind for the test ?

As per having a good reliable url, I first thought using a local instance of ffserver would be a good idea but discovered that it did not support as many protocols as ffmpeg is able to play. Anyway, it might still be a reliable way to reproduce some codec-related issues if it can be launched with different parameters for several tests.

Issue #45, opened 2015-11-09
Changing audio/video properties mid stream


nickkelsey:
Sometimes digital TV from an antenna (live or recorded) change key properties mid stream.

Audio might switch from 5.1 to 2 channel between a show and advertising, or the video might switch from 1080i (1920x1080x30fps) to 720p (1280x720x60fps) between shows.

The audio path uses swr_convert() to convert from floating point to S16 samples - this API segfaults internally when audio changes from 5.1 to 2 channel. I fixed the interop code to reinitialize the swr_convert context to 2/5.1 channel as needed - this fixes the segfault but now I need to tell the MediaFoundation layer that the PCM data I am passing it is now only 2 channel without interrupting playback.

Can you suggest a way to update the MediaFoundation properties mid-stream without interrupting playback?

(Currently focused on the audio change. Guessing the video size/framerate changing mid stream will have a similar solution).

Nick

nograx:
Any news on this?

reego-fr:
Hi,
I don't think the channel count can be changed during playback on the MediaStreamSource api. Can @khouzam confirm ? 
If it is true I see two workarounds : 

- create a new media stream source when channel count change,
- or use ffmpeg resampler to convert 2.0 to 5.1, preferred in your case. Remember that if you start playing the stream as 2.0 you won't be able to upgrade it to 5.1, so you might want to hardcode 5.1 output or more instead of letting the playback be initialized as stereo.

reego-fr:
I fixed it using ffmpeg resampler but I think the first option would be the only way to do it when you don't know how many channels to expect when starting playback.

nickkelsey:
Confirming, a reasonable work around for FFmpeg decoded audio is to detect the number of audio channels the defaulter audio render has (ie how many actual speakers there are), then use the FFmpeg resampler to convert the number of channels. This trick doesn't work when using the a Media Framework codec like AAC and resampling video is just nasty... still needs a real fix.

reego-fr:
Thanks for the feedback, I didn't know about related issues with aac, but I found AC3 decoder from Media Framework was behaving pretty much like the resampler fix. 
I forgot you mentioned video as well ! Nasty is just right :-)

OliverS79:
@reego-fr - I like your proposal. Can you share the code you are using to leverage ffmpeg's resampler?

nograx:
@reego-fr - A code example would be nice. Thanks.

reego-fr:
[...]

PR #47, opened 2015-12-01
fix MPEG-PS incorrect packet position

senkia:
It seems that in the MediaSampleProvider.GetNextSample method when we get the current AVPacket, the packet's presentation timestamp is incorrect (it's actually larger that the duration of the movie I'd like to play). This timestamp should be 0 at the very first packet but it isn't.

Fortunately this error is consistent meaning that each AVPacket presentation timestamp is larger than it should be by the first packet's timestamp (because that one's should be 0). So if I save this constant "offset" (StartPosition attribute) at the first packet (the offset equals with the first packet's timestamp) and subtract it from every packet's presentation timestamp the video works just fine with seek and everything.

### DIFF
--- a/FFmpegInterop/Source/MediaSampleProvider.cpp
+++ b/FFmpegInterop/Source/MediaSampleProvider.cpp
@@ -23,6 +23,8 @@
 
 using namespace FFmpegInterop;
 
+int64 StartOffset = 0;
+
 MediaSampleProvider::MediaSampleProvider(
 	FFmpegReader^ reader,
 	AVFormatContext* avFormatCtx,
@@ -32,19 +34,27 @@ MediaSampleProvider::MediaSampleProvider(
 	, m_pAvCodecCtx(avCodecCtx)
 	, m_streamIndex(AVERROR_STREAM_NOT_FOUND)
 {
+	DebugMessage(L"MediaSampleProvider\n");
+	StartOffset = 0;
+
 }
 
 HRESULT MediaSampleProvider::AllocateResources()
 {
+	DebugMessage(L"AllocateResources\n");
+	StartOffset = 0;
 	return S_OK;
 }
 
 MediaSampleProvider::~MediaSampleProvider()
 {
+	StartOffset = 0;
+	DebugMessage(L"~MediaSampleProvider\n");
 }
 
 void MediaSampleProvider::SetCurrentStreamIndex(int streamIndex)
 {
+	DebugMessage(L"SetCurrentStreamIndex\n");
 	if (m_pAvCodecCtx != nullptr && m_pAvFormatCtx->nb_streams > (unsigned int)streamIndex)
 	{
 		m_streamIndex = streamIndex;
@@ -54,6 +64,7 @@ void MediaSampleProvider::SetCurrentStreamIndex(int streamIndex)
 		m_streamIndex = AVERROR_STREAM_NOT_FOUND;
 	}
 }
+
 
 MediaStreamSample^ MediaSampleProvider::GetNextSample()
 {
@@ -100,7 +111,15 @@ MediaStreamSample^ MediaSampleProvider::GetNextSample()
 		// Write the packet out
 		hr = WriteAVPacketToStream(dataWriter, &avPacket);
 
-		Windows::Foundation::TimeSpan pts = { LONGLONG(av_q2d(m_pAvFormatCtx->streams[m_streamIndex]->time_base) * 10000000 * avPacket.pts) };
+		if (StartOffset == 0) {
+			//if we havent set StartOffset already
+			DebugMessage(L"Saving startoffset\n");
+			
+			//in some real-time streams avPacket.pts is less than 0 so we need to make sure StartOffset is never negative
+			StartOffset = avPacket.pts < 0 ? 0 : avPacket.pts;
+		}
+
+		Windows::Foundation::TimeSpan pts = { LONGLONG(av_q2d(m_pAvFormatCtx->streams[m_streamIndex]->time_base) * 10000000 * (avPacket.pts - StartOffset)) };
 		Windows::Foundation::TimeSpan dur = { LONGLONG(av_q2d(m_pAvFormatCtx->streams[m_streamIndex]->time_base) * 10000000 * avPacket.duration) };
 
 		sample = MediaStreamSample::CreateFromBuffer(dataWriter->DetachBuffer(), pts);
@@ -154,11 +173,9 @@ AVPacket MediaSampleProvider::PopPacket()
 
 void MediaSampleProvider::Flush()
 {
+	DebugMessage(L"Flush\n");
 	while (!m_packetQueue.empty())
 	{
 		av_free_packet(&PopPacket());
 	}
 }
-
-
-


reego-fr:
Hi,

I just looked at the code without testing it but I am wondering if there is a risk that audio/video shift happens using this fix. 
I remember some files that would probably fail if I'm right. I will report about it.

reego-fr:
I can confirm what I said above. An obvious audio/video sync issue appeared on one of my — weird — sample files. 
Your approach is still valuable as opposed to using ffmpeg's logic to get the start offset but you have to look at the several streams as a whole before applying the offset. Does it make sense to you ?

senkia:
Yes like I said before it's definitely not the best solution. I'm using this in an app that only has to open the files recorded by the same IP camera so I don't have to handle multiple formats but still after some seeking or play/pause switching I also run into problems.

michaelmayr:
Hi,
I tried to correct this problem by using the value from ```AVFormatContext::start_time``` as offset for pts calculation in MediaSampleProvider. With this solution both, audio and video will have the same start point and audio/video sync issues should not appear.

See also 94e6d58ff85323c5fdb334b140b5809c87e081db, tested with RTP streams from an IPTV provider.

reego-fr:
Hi,
you're right. For a complete solution you can also apply the same offset when seeking.

mmayr-at:
Thank you for pointing this out. I haven't considered seeking yet, as I was working only with live streams so far.

yodurr:
please update to complete solution and we'll revisit taking the change

mmayr-at:
Please see commit 02b3be1ef04a988130f91873763616513e9f4c84 and  94e6d58ff85323c5fdb334b140b5809c87e081db, if you mean my solution.

alakshmi1030:
Hello @senkia, is the pull request updated with the complete solution involving the offset while seeking? We want to take this pull req in but we want to make sure there are no other issues and the other concerns have been addressed regarding @reego-fr earlier comment on the complete solution.

senkia:
Hi
No my solution was not complete, please see the solution linked by @mmayr-at that's the one I've ended up using myself.

reego-fr:
Hi,
@mmayr-at solution is good. 
I added a small improvement to allow some more files to play (these still fail when seeking but ffmpeg does not support these files either). Please see https://github.com/Microsoft/FFmpegInterop/commit/2ed49d4b1b7d31b6b9203bb552d955cae897e1f6

mitchcapper:
@reego-fr  what do you mean by ffmpeg does not support seeking those files? What types of files are you talking?

reego-fr:
Hi, 
here is a sample file to demonstrate the seek issue and show how my fix allows such files to start playing : H265 tears of steal sample on https://x265.com/hevc-video-files/
I don't know exactly why ffmpeg does not seek within this file but ffplay itself fails to do it too (playback restarts from the beginning).

mitchcapper:
What is interesting is apps like VLC (using ffmpeg) can seek it though:)

khouzam:
It looks like av_seek_frame is failing for the x265 video. Not sure why though.

mitchcapper:
it also fails on streaming (like HLS) for example

reego-fr:
I had better results using AVSEEK_FLAG_BACKWARD (as does VLC) but it is hard to understand why. 
My previous fix is not even needed anymore, the file will start playing correctly. 
I will have to try other files as well to check for regressions because my previous attempts to change the seek method failed. 
The same sample shows a significant delay when changing the position, longer than when played on VLC. I am not sure about what may be causing this.

reego-fr:
@khouzam, I had no issue with other h265 samples. Did you ?
I wonder if this sample just don't have enough keyframes to be seekable. It might be ignored as well !

reego-fr:
After av_seek_frame with AVSEEK_FLAG_BACKWARD playback restarts from the beginning. It explains the delay experienced on displayed frames.

khouzam:
@reego-fr 

I only tried a short x265 sample, didn't dig into it at all. Could be that there aren't enough keyframes, but then I would expect it to pickup the previous one and discard frames instead of the next one. But then, that might be what the AVSEEK_FLAG_BACKWARD does.

reego-fr:
Yes. It appears that ffmpeg is not discarding the frames by itself, at least on the version I am testing with. Other players relying on it must do it in their code.

khouzam:
I wonder if that could explain the freeze when seeking with uncompressed samples

reego-fr:
Hi,
unfortunately I could not find another workaround than setting the buffer size to 0 to solve the seek issue with uncompressed samples. 
Reporting a timestamp that matches the following samples on OnStarting event helped with the short stream above but not in general. 
Now that I tested h265 I noticed some stuttering inside some high res videos. The cpu looks high but ok. I wonder what might be causing this but I think a buffer would probably help hiding it :-) Is there any chance we can use the one from the media stream source ?

reego-fr:
As far as I am concerned, https://github.com/Microsoft/FFmpegInterop/commit/02b3be1ef04a988130f91873763616513e9f4c84 is good to solve the issue. My commit still enhance compatibility without changing the seek method. Seek can be fine tuned a bit but I think it is a separate issue. 
I didn't look at the HLS issue specifically though. @mitchcapper do you have any sample stream to share ?

mitchcapper:
Sure,  so two samples that work just fine in VLC but can struggle with this
one a straight up file:
http://mitchcapper.com/vid/4013_20160503141000.ts   seeking will sometimes work, will sometimes jump to the end with an invalid timeline length. Choppy audio.

or an HLS: http://mitchcapper.com/vid/4013_20160503141000.ts.640x368_800kV_64kA.m3u8

The HLS doesn't have the jump to the end issue on fd353fe but it can freeze when seeking and not resume playback.  I have looked at some of the buffer composition code to see if its related to how the buffer is being filled.

If smaller samples would be helpful let me know.

reego-fr:
HLS definitely needs a buffer. I don't know if ffmpeg has one but I got better results when leaving enabled the one from the media stream source. Playback pauses frequently without it. 
Your sample didn't play well on my side too.

reego-fr:
Hi @mitchcapper,
I found your file had three more issues : 
- Choppy audio is caused by a channels count change shortly after the beginning. Someone else experienced this as well here https://github.com/Microsoft/FFmpegInterop/issues/45
- Right after the beginning too, there must be corrupted or missing frames (video, or both audio + video) that freezes video playback. 
- With compressed samples playback, the video sample have an invalid pts. I don't remember seing this with other ts files.

mitchcapper:
I assume you refer to the non HLS, h264 stream (as the HLS stream should be fairly stable as its re-encoded by ffmpeg).
Certainly the samples are not ideal,  (taken from an HDPVR capture box) unfortunately we cannot always control the content we have to play, and native ffmpeg /vlc handles these sorts of files without issue.   Is it the corruption that causes the stream to sometimes jump to the end with an invalid time offset or the channel count change?  I am a bit more curious on this as I have seen similar behavior with files from other sources (where it jumps to the end of the recording).

reego-fr:
Hi @mitchcapper,
sorry for the delay. 
I quickly tried a really bad DVB-T rip of my own and found out the media foundation h264 decoder did its job until the end of the file, one minute or so. It looks solid enough to me for now. What happens with your file is something else. 
To answer your question, when the player position jumps to the end it is very likely that is it playing a sample with that timestamp. It turns out half of the samples in your file have a really big one because they have none ! No timestamp translates to a big value. 
The channel count change looks common on mpeg-ts ac3 streams but I don't know if it is documented anywhere. It only make sense during commercials or between programs which does not look to be the case in your file. 
Your file also has what looks like a timestamp reset, I think it is causing video to stop. 
The last two issues are the two I know about mpeg-ts that would cause issues with the current code. What bugs me is that you have them both in your file. Is it the result of a channel switch ? I tried and it didn't look really nice obviously. 
No matter how you got this file, the good news is that it mostly suffers from known issues. It would play nice with some fixes or workarounds.

mitchcapper:
thanks @reego-fr for the explanation.   The Media Foundation support over ffmpeg does seem better in many cases, although I was looking to use ffmpeg instead of MF.  In addition MF can run into a few of the same issues as FFMPEG (like the timestamp problem) that VLC avoids.
There was no channel change during the recording, but maybe one was made shortly before hand and the HDPVR only adjusts audio at increments (I am not sure).  Of course it could be the source program itself changing, but why a news program would do so (Unless it has some cut together clips) I am not sure.  I just quickly recorded a clip from live tv to use, it is by no means a specific example that had a lot of issues ( but it seems it does ).  The timestamp issue is more curious, as I see that happen on various files from a few sources so a pointer on where to look for those sort of fixes would be great to hear so I can see about implementing them for playback.

reego-fr:
I don't understand the reason why you prefer to use ffmpeg over Media Foundation. I assume you refer to uncompressed sample playback vs. compressed. Did you consider using VLC instead, I mean libvlc ? 
I don't remember seing something explicitly related to a timestamp reset in ffplay source code. It should handle this case though. 
I wonder if the media duration retrieved from ffmpeg will be correct for these files, if it is the fix shouldn't be to hard to implement. I never had a chance to look at one of those files but I think yours does not represent the general case and another sample would be useful for the tests.

mitchcapper:
Ah, MF is good but it has a few problems:
*) FFmpeg can play more formats than MF, while a try MF first then FFmpeg is totally possible in the case of failure, sometimes MF just plays files worse (although currently it is often the reverse).  There would be no easy way to detect 'worse' playback of formats.  For example the timeline bug which sometimes occurs with MF and there is likely no solution to for MF due to limited access to enhancing it.  It is something VLC is able to work around with FFmpeg however.   There are other times (just in experience of VLC vs MF) where multi-channel audio DTS issues arise or similar playback problems so using FFmpeg is good without having to have a user say 'this is playing crappy switch to ffmpeg".

*) MF is somewhat limited in advanced actions you can do, ffmpeg however has a much wider range of parameters and options that it can take.  In addition FFmpeg has more function options for example seek to a byte position.

libVLC is a possible option, it adds another layer into the mix though, so if the items are resolvable more at the ffmpeg level it has advantages.  FFmpeg is generally first and fastest to fix bugs (MF however is probably the slowest) with others requiring waiting for it to propagate up.  So the ability to have a library built right on top of FFmpeg that was fairly robust has several advantages.

I will be happy to pull some more tests.  What I can say is MF struggles (often differently than ffmpeg) with more than just items encoded by the hdpvr (what the previous samples were) including some raw OTA recordings.   Its OTA unlike the HDPVR so there ends up being some corruption in the files.

reego-fr:
I think MF deinterlacing is the best compared to others. It can probably be enabled on uncompressed samples playback but I will stick to compressed samples playback because it applies it automatically :-) The media stream source api has some limitations and issues, so does ffmpeginterop. Hopefully there are things that can be fixed.

mitchcapper:
Agreed, although MF probably moves the slowest to fix the issues, thus the benefit of a 3rd party library.

reego-fr:
Sure, but you will still be using MF pipeline in the end :-)
Anyway, it means there are two issues instead of three that prevent your file, amongst others, to play fine as uncompressed samples. This is good news !

nograx:
Hi, could you explain how to use AVSEEK_FLAG_BACKWARD? Want to test it in my scenario as I also have problems with seeking. Thanks

reego-fr:
Hi @nograx, AVSEEK_FLAG_BACKWARD is used with [av_seek_frame](http://ffmpeg.org/doxygen/trunk/group__lavf__decoding.html#gaa23f7619d8d4ea0857065d9979c75ac8).

reego-fr:
Back to the main issue, should we open a new pull request with the updated code ?
Please note I have another fix for mpeg-ts pts rollover, but the commits above are still good because the original issue is about mpeg-ps :-)

reego-fr:
I found my commit to fix the short x265 sample was wrong. https://github.com/Microsoft/FFmpegInterop/commit/02b3be1ef04a988130f91873763616513e9f4c84 is still the solution, needing another little tweak to restore playback of the same short sample. I don't think it has other side effects.

nograx:
I have good results using AVSEEK_FLAG_FRAME for recorded movies from my dvb-s receiver. I can seek without problems and the files are playing nice.

I have just one more problem, already descripted above. If I seek in some files, audio is choppy and the file is played with about 1/4 of the normal rate. Audio is A52 Audio (aka AC3) (a52) in this case.

reego-fr:
Hi @nograx, what are the issues you encountered when using no flag instead of AVSEEK_FLAG_FRAME ? I would like to know what are the flaws of the default seek implementation. 

Regarding the audio issue, there are 99% chances it is caused by a channel count change indeed. Fortunately, it is repeatable and easy to debug. 
I am not sure what is the best fix yet, because the channel count is set on MF beforehand and probably can't be changed, but a workaround can be implemented quickly.

reego-fr:
As a side note, I know MF provides a AC3 decoder. I wonder if the channel count change is supported in it. 
Do the guys at Microsoft know if it is, and what prevented ac3 to be included in the compressed sample formats supported by FFmpegInterop ? I know for experience it plays my streams but I could not get it to work in this project for an unknown reason when I tried a long time ago.

reego-fr:
Actually I tried the AC3 decoder. It looks like it works but will not switch from 2 to 6 channels automatically.

nograx:
I have 3 scenarios in my app:

1. Stream Live-TV over http, seeking not necessary
2. **Stream** Recordings over http:
    - Default Seek Implementation: When Seeking Video **Stops playing**
    - AVSEEK_FLAG_FRAME: **Seeking is possible** without problems
3. **Downloaded** recordings (played locally):
    - Default Seek Implementation: When Seeking Video **Stops playing**
    - AVSEEK_FLAG_FRAME: When Seeking Video **Stops playing**

So I don't know why I can seek with AVSEEK_FLAG_FRAME when I stream the file and the video stops playing when I seek with AVSEEK_FLAG_FRAME on downloaded file...

My workaround for AC3 issue: I'm trying to play the video without FFMPEG. I will then check for MediaFailed Event and if it's an decoding issue, I will play again with FFMPEG (because files with AC3 can be played without FFMPEG).. Not nice but it works so far...

reego-fr:
Given your results I wonder if you know about the media stream source buffer issue described above. 
I didn't expected AVSEEK_FLAG_FRAME to make a difference, would you have sample code to share ? Also, what is the protocol used to stream your recordings over http and what are the codecs ? 

Anyway, I don't think it belongs here, so are most of the posts here. Correct me if I am wrong, but I want to make it clear for people trying to sort it out. 
This pull request is not that much about seeking but about the playable range. The proposed fix is not what causes most of the issues discussed above and only aims at fixing one seeking issue.

reego-fr:
Hi,
here is my update on this issue: https://github.com/reego-fr/FFmpegInterop/commit/6a6c196fdf48bbf1c842eeb26bfad009e38c24ee
According to my tests it fixes the start time issue as it should and does not create new issues with files that played on https://github.com/Microsoft/FFmpegInterop/commit/6649f4b942675adfd72c0f38471de56f51c43436

OliverS79:
@reego-fr: Thanks for the fix. I can confirm 6a6c196 fixed my issues I had with playing some MPEGTS streams. With this, video and audio is always in sync. Thanks a lot.

@khouzam: Any chance this entire branch becomes part of the official release?

FFmpegInterop/Source/MediaSampleProvider.cpp
```diff
@@ -100,7 +111,15 @@ MediaStreamSample^ MediaSampleProvider::GetNextSample()
 		// Write the packet out
 		hr = WriteAVPacketToStream(dataWriter, &avPacket);
 
-		Windows::Foundation::TimeSpan pts = { LONGLONG(av_q2d(m_pAvFormatCtx->streams[m_streamIndex]->time_base) * 10000000 * avPacket.pts) };
+		if (StartOffset == 0) {
```
khouzam: Won't this cause the PTS to be off by one frame? If the start offset is actually 0, then we'll set the StartOffset as the next timestamp found. Should StartOffset be initially set to a negative value to allow for the value of 0 to be valid?

reego-fr:
@khouzam I think @OliverS79 was mentioning https://github.com/Microsoft/FFmpegInterop/commit/6a6c196fdf48bbf1c842eeb26bfad009e38c24ee branch

OliverS79:
Yes, @reego-fr and @khouzam - I was refering to the reego-fr@6a6c196 branch; which should not include the line of code you refer to. (if (StartOffset == 0)...). This (reego-fr@6a6c196) seams to be the best fix for the problem.

nograx:
Do you have a solution for the channel count change mid stream already or any idea for a workaround? I really need a fix for it.

OliverS79:
@nograx @reego-fr - I don't have a solution for the channel switch problem, however I think I found the reason why AC3 is always send to ffmpeg and not directly to MF. While MF on the desktop supports AC3, it does not support it on Windows Mobile 10. Speaking of Win10 Mobile. Has one of you managed to play a MPEG2TS with MPEG2 video and AC3 audio without audio stutter on Win 10 Mobile? My program runs perfect on the Desktop, but the combination of MPEG2 video and AC3 audio is causing poor audio quality on my Lumia 950. H264 and AC3 plays perfectly on the phone.

reego-fr:
@nograx, I have updated https://github.com/Microsoft/FFmpegInterop/issues/45 on the channel count change issue.  
@OliverS79 , mpeg2 and ac3 should be ok. I only tried low res video though and don't know how much software decoding can handle. I can't say about Win 10 mobile because I am still on 8.1 only.

khouzam:
Thank you. Finally merged. :-)

reego-fr:
Hi @khouzam,
I reviewed the merge and I think the wrong commit was merged. I don't remember the whole story because I worked on that a long time ago but you should take my commit or @mmayr-at 's latest if AVSEEK_FLAG_BACKWARD was added in the meantime which I think is the case given the history I read.

reego-fr:
https://github.com/Microsoft/FFmpegInterop/commit/a8fe94fbfc1c908e55bcbfedbbc706d34a4b25f7 adds other issues and is not the proper way of dealing with the original issue.

mcosmin222:
This PR needs to be reverted.  See #172. There has to be a smarter way to determine the starting offset.

Issue #64, opened 2016-03-11
failed to build FFmpeg for winRT in win10 x64 with VS2015 installed


konankid:
Hello expert,
I am trying to use FFmpegInterop for win10 UWP application.
So I following the guidance "build FFmpeg for WinRT" https://trac.ffmpeg.org/wiki/CompilationGuide/WinRT trying to compiler for Windows 10 x64:
here is some environment information:
VS2015 Enterprise with 10.0.240.0 SDK
OS: Windows 10 Enterprise x64
PATH were set to:Program Files (x86)/Microsoft Visual Studio 14.0/VC/BIN/x86_ARM

so when I check:
$ which cl
/c/Program Files (x86)/Microsoft Visual Studio 14.0/VC/BIN/x86_ARM/cl

$ which link
/c/Program Files (x86)/Microsoft Visual Studio 14.0/VC/BIN/x86_ARM/link
 
But when I perform "BuildFFmpeg.bat win10" I got one error:
"cl is unable to create an executable file.
C compiler test failed."

Am I configure anything wrong? Is anybody have the same environment with me and get green build?
Or alliteratively, where can I get binary for that?

khouzam:
Hi, can you look at the configure log to see what failed, and perhaps attach it?

konankid:
# ## Hi khouzam,

Last lines of my config.log file:

> WARNING: pkg-config not found, library detection may fail.
> mktemp -u XXXXXX
> nV13AC
> check_ld cc
> check_cc
> BEGIN ./ffconf.IeVPbA8A.c
>     1	int main(void){ return 0; }
> END ./ffconf.IeVPbA8A.c
> cl -nologo -D_USE_MATH_DEFINES -D_CRT_SECURE_NO_WARNINGS -D_CRT_NONSTDC_NO_WARNINGS -MD -DWINAPI_FAMILY=WINAPI_FAMILY_APP -D_WIN32_WINNT=0x0A00 -c -Fo./ffconf.PMS0DaMt.o ./ffconf.IeVPbA8A.c
> ffconf.IeVPbA8A.c
> /z/FFmpegInterop-master/ffmpeg/compat/windows/mslink -APPCONTAINER WindowsApp.lib -nologo -out:./ffconf.0ZsG94EB.exe ./ffconf.PMS0DaMt.o
> LINK : fatal error LNK1181: cannot open input file 'WindowsApp.lib'
> C compiler test failed.
> 

To get this I've run below command:

[...]

Issue #49, opened 2015-12-08
Cannot run config file 


bilal-sou:
Hi 
Following the compilation instructions I've installed verything (msys2 yasm etc..), set up the environment variables, checked with "which" where the cl command was pointing at, check my PATH but nothing worked. I keep on getting 

WARNING: pkg-config not found, library detection may fail.
mktemp -u XXXXXX
ZYw84V
WARNING: Unknown C compiler cl, unable to select optimal CFLAGS
check_ld cc
check_cc
BEGIN ./ffconf.oc8wxDOw.c
    1	int main(void){ return 0; }
END ./ffconf.oc8wxDOw.c
cl -MD -DWINAPI_FAMILY=WINAPI_FAMILY_PC_APP -D_WIN32_WINNT=0x0603 -c -o ./ffconf.T0zVIMc6.o ./ffconf.oc8wxDOw.c
../../../configure: line 822: cl: command not found
C compiler test failed.

I just want to compile ffmpeg without even the interop part as I don't really need it. Could you please give me any hint on what could be going on ?

Thanks

timotiusmargo:
Before you run configure in your MSYS2 shell, what is the output of `which cl` in the same shell?

Could you also share your Visual Studio and Windows 10 SDK/tools version?

Issue #40, opened 2015-10-08
Problem running config..


Giulio-Ladu:
Hi there,
I am currently on the section labelled
Windows Store 8.1 x64 (Windows 8.1 x64 in Visual Studio) 
After creating the directories I have the following : FFmpegInterop>ffmpeg>Output>Windows8.1>x64
When I try the following:
$ ../../../configure \
> --toolchain=msvc \
> --disable-programs \
> --disable-dxva2 \
> --arch=x86_64 \
> --enable-shared \
> --enable-cross-compile \
> --target-os=win32 \
> --extra-cflags="-MD -DWINAPI_FAMILY=WINAPI_FAMILY_PC_APP -D_WIN32_WINNT=0x0603" \
> --extra-ldflags="-APPCONTAINER" \
> --prefix=../../../Build/Windows8.1/x64
I get the following error: 
c99wrap cl is unable to create an executable file.
C compiler test failed.

If you think configure made a mistake, make sure you are using the latest
version from Git.  If the latest version fails, report the problem to the
ffmpeg-user@ffmpeg.org mailing list or IRC #ffmpeg on irc.freenode.net.
Include the log file "config.log" produced by configure as this will help
solve the problem.

I have tried updating the ffmpeg file using the following git clone git://source.ffmpeg.org/ffmpeg.git
but I still get the same error? 
Is this something I am doing wrong or is there currently a problem?
Thanks

khouzam:
Have you tried the buildffmpeg.bat script? It will pass the right parameters to build FFmpeg if your using the submodule in the project.

Also if you can look at config.log that would help figure out what's failing.

Thanks

Giulio-Ladu:
Hi  there, 
Thanks for quick reply!
I did try the batch file, and got the same problem..

this is the problem from the config.log

WARNING: Unknown C compiler c99wrap cl, unable to select optimal CFLAGS
check_ld cc
check_cc
BEGIN ./ffconf.0J5ekFjp.c
    1	int main(void){ return 0; }
END ./ffconf.0J5ekFjp.c
c99wrap cl -MD -DWINAPI_FAMILY=WINAPI_FAMILY_PC_APP -D_WIN32_WINNT=0x0603 -c -o ./ffconf.HhEkGEuc.o ./ffconf.0J5ekFjp.c
../../../configure: line 841: c99wrap: command not found
C compiler test failed.

I am not too great with C++, so do not know if this is something I can fix, or if it is a config error
[...]

PR #73, opened 2016-05-12
export full current PATH from environment into MSYS2

stammen:
This pull request fixes the cl.exe not found error when building ffmpeg for winrt with newer versions of msys. We need to call set MSYS2_PATH_TYPE=inherit from BuildFFmpeg.bat since the script directly launches bash.exe.

This PR fixes issues:

https://github.com/Microsoft/FFmpegInterop/issues/64
https://github.com/Microsoft/FFmpegInterop/issues/60
https://github.com/Microsoft/FFmpegInterop/issues/40
https://github.com/Microsoft/FFmpegInterop/issues/49

### DIFF
--- a/BuildFFmpeg.bat
+++ b/BuildFFmpeg.bat
@@ -12,6 +12,9 @@ set BUILD.win10=N
 set BUILD.win8.1=N
 set BUILD.phone8.1=N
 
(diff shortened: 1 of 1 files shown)


khouzam:
Thanks Dale

PR #82, opened 2016-07-28
Fix for compressed HE-AAC + HE-AAC v2 and uncompressed HE-AAC v2

reego-fr:
Hi,

here are fixes for HE-AAC and HE-AAC v2. 
Some test files can be found at http://lists.apple.com/archives/streaming-server-users/2005/Aug/msg00007.html. 
ps.3gp and sbr.3gp (by http) are the files you will want to try but be advised and first lower down the volume because the tones may be unpleasant (10kHz).


reego-fr:
To clarify a bit, the second commit resolves an issue with HE-AAC v2 embedded in ADTS that was introduced by the first one. It looks like the stream properties are not reported the same way by ffmpeg in this case.

reego-fr:
ADTS can be tested against AAC shoutcasts, HE-AAC v2 should be leveraged on low bandwidth streams. These should be easy to find but I can send you a link if you need it.

lukasf:
This is pretty old, still, some thoughts from my side.

What you are actually doing is downgrade AAC-HE and AAC-HEv2 to AAC-LE. AAC-HE stores extra data to double the sample rate. AAC-HEv2 stores additional data to double channels form mono to stereo. By forcing half sample rate and channel count, you force a downgrade to the legacy AAC-LE stream.

I think the problem with ffmpeg is that it decodes the first few frames with the lower sample rate and channel count. This is probably only be the case with streaming data, where the extra data is not there during first few frames. Later, ffmpeg switches to a higher sample rate and channel count, once extra data has been detected. But since you force the resampler to the lower sample rate and channel count, that information is lost. You will end up with 24khz stereo for AAC-HE and 24khz mono for AAC-HEv2, although both are actually stereo 48khz. While it makes the streams at least playable, it only a workaround and does not address the underlying problem.

Refer this:

https://video.stackexchange.com/questions/17677/why-does-aac-have-two-sampling-rates

Maybe it would be better to try to pass through AAC to Windows codecs. They support all the formats and do the proper decoding and upgrade to higher sample rate / channel count. Alternatively, we could try to notify the MSS that encoding properties have been changed. I have not tried this, but I think you can change the actual values in EncodingProperties before sending a frame, if you detect a change in sample rate or channel count. At least this is how I understand the msdn documentation. Any change would also require the resampler to be recreated in UncompressedAudioSampleProvider.

reego-fr:
Thanks for your inputs @lukasf 

I think I will have to double check this because I have not put my hands on it for a long time. 
One thing I am sure about is that I would not consider the playback problem to be fixed when only rendering half of the media, either regarding the sample rate or regarding the channel count. While one can easily be mistaken on the former, checking whether the output is mono or stereo sounds like a simple task with no room for testing errors. 
I would be surprised if I ended up downgrading the stream without explicitly telling about it, but maybe there is something I overlooked, or the AAC Windows Codec changed its behaviour in the meantime. 
I though I included a link to an old documentation, apparently not here but in an issue instead, where it was stated how sample rates and channel count should be declared when asking it to decode AAC-HE and AAC-HEv2. At that time what the doc said and what I observed during testing was consistent.

lukasf:
Please forget what I wrote. I think I confused your "passthrough" declaration with the "force decode" declaration. When looking into the sources again, I think you are doing it the right way. Sorry for jumping the gun too quickly. 

Now I only have a doubt about channel declarations when decoding with ffmpeg. In the audio descriptor, you always set to channels * 2 for v2. But in the resampler, you only set to channels * 2 if extradata is available in the context. This looks kind of inconsistent.

reego-fr:
I can't remember right now why I did it this way, I should be able take a look and come back to you in a few days.

reego-fr:
You were right @lukasf 
It looks like the fix was not complete. It should be a lot better now. Thanks.

lukasf:
You'll need to merge latest changes from master branch...

reego-fr:
Interesting. Different formats apparently report the channel count differently. 
I like your solution better than enumerating all the formats and see what they report. 
I wonder if more than two channels are possible but I already thought about hardcoding it to two so that it causes less troubles. 
I'll double check that and try to see if there is a better way of getting the channel count from ffmpeg.

lukasf:
The definition of HE-AAC V2 is "HE-AAC plus Parametric Stereo". So I think HE-AAC V2 is always 2 channels.

PR #83, opened 2016-08-01
Enable uncompressed video deinterlacing

reego-fr:
Hi,
here is an update to apply deinterlacing on uncompressed video frames. 
It worked well for me so far.

### DIFF
--- a/FFmpegInterop/Source/FFmpegInteropMSS.cpp
+++ b/FFmpegInterop/Source/FFmpegInteropMSS.cpp
@@ -24,6 +24,7 @@
 #include "UncompressedAudioSampleProvider.h"
 #include "UncompressedVideoSampleProvider.h"
 #include "shcore.h"
+#include <mfapi.h>
 
 extern "C"
 {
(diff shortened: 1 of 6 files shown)


reego-fr:
I am not sure if we have to follow the rules explained at https://msdn.microsoft.com/en-us/library/windows/desktop/aa367729(v=vs.85).aspx but you will notice I am only setting the interlaced attributes on progressive frames, not the other two and it looks ok. I find it strange to have to specify them but if it has to be added it is certainly not a lot of work to do so.

khouzam:
This is great, thanks @reego-fr 

Is there a piece of interlaced content that can be added to the unit tests to validate this functionality?

reego-fr:
I am not aware of any publically available stream over the internet being interlaced even though it should help reduce the bandwidth requirements in some cases :-) 
Maybe a file downloaded from http would do it ? It didn't work with the site I downloaded samples from but it might work with other servers. The only thing I am concerned about is respecting terms and conditions of these sites which could not allow such use of their service.

OliverS79:
@reego-fr - Just want to let you know that your piece of code works well for me! It de-interlaces DVB content perfectly. Thanks a lot for this - like all you other stuff you developed. Good work!
@khouzam - like for the time fix issue, it would be great if also this functionality can make it in the official release at it eases updates in the future.

khouzam:
Thanks @reego-fr

Issue #80, opened 2016-06-28
Build Error : FFmpeg 3.1v


dingpong:
Build Errors

1>src\ffmpeg\ffmpeginterop\source\ffmpeginteropmss.cpp(287): error C4996: 'AVStream::codec': was declared deprecated
1>src\ffmpeg\ffmpeg\build\windows10\x64\include\libavformat\avformat.h(880): note: see declaration of 'AVStream::codec'
1>src\ffmpeg\ffmpeginterop\source\ffmpeginteropmss.cpp(325): error C4996: 'AVStream::codec': was declared deprecated
1>src\ffmpeg\ffmpeg\build\windows10\x64\include\libavformat\avformat.h(880): note: see declaration of 'AVStream::codec'
1>src\ffmpeg\ffmpeginterop\source\uncompressedaudiosampleprovider.cpp(109): error C4996: 'avcodec_decode_audio4': was declared deprecated
1>src\ffmpeg\ffmpeg\build\windows10\x64\include\libavcodec\avcodec.h(4704): note: see declaration of 'avcodec_decode_audio4'
1>src\ffmpeg\ffmpeginterop\source\uncompressedvideosampleprovider.cpp(123): error C4996: 'avcodec_decode_video2': was declared deprecated
1>src\ffmpeg\ffmpeg\build\windows10\x64\include\libavcodec\avcodec.h(4753): note: see declaration of 'avcodec_decode_video2'

dingpong:
I had to temporarily modify three functions.
[ffmpeg3_1v.txt](https://github.com/Microsoft/FFmpegInterop/files/336423/ffmpeg3_1v.txt)

khouzam:
I've created a new branch, `helmut-steiner-ffmpeg3.1`, which has @helmut-steiner changes to incorporate FFmpeg 3.1 as well as the changes in logic for the new `avcodec_send_packet` and `avcodec_receive_frame` logic. I've tried it on a few files and things look good. According to the samples that we pull out of the file and the ones that we pull out of the decoder, everything matches up.

I need to do some more testing with a bigger variety of files, but I'd like to see if other can try it and see if they find any issues.

Thanks

PR #88, opened 2016-09-05
ffmpeg updated to v3.1 and resolved build errors in ffmpeginterop

helmut-steiner:
According to issue #80 - Build Error : FFmpeg 3.1v 
Incorporated fixes from dingpong

### DIFF
--- a/FFmpegInterop/Source/FFmpegInteropMSS.cpp
+++ b/FFmpegInterop/Source/FFmpegInteropMSS.cpp
@@ -284,7 +284,23 @@ HRESULT FFmpegInteropMSS::InitFFmpegContext(bool forceAudioDecode, bool forceVid
 		audioStreamIndex = av_find_best_stream(avFormatCtx, AVMEDIA_TYPE_AUDIO, -1, -1, &avAudioCodec, 0);
 		if (audioStreamIndex != AVERROR_STREAM_NOT_FOUND && avAudioCodec)
 		{
-			avAudioCodecCtx = avFormatCtx->streams[audioStreamIndex]->codec;
+			/** allocate a new decoding context */
+			avAudioCodecCtx = avcodec_alloc_context3(avAudioCodec);
+			if (!avAudioCodecCtx) {
+				fprintf(stderr, "Could not allocate a decoding context\n");
+				avformat_close_input(&avFormatCtx);
+				return AVERROR(ENOMEM);
+				
+			}
+			
+			 /** initialize the stream parameters with demuxer information */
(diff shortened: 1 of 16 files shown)


khouzam:
Hi @helmut-steiner 

Thanks you very much for the contribution,

I'm taking a look at the code and there are few things that I'm looking at fixing before accepting the pull request, notably the coding style to stay in line with the rest of the project as well as the early returns, it looks as though the failures would result in some memory leaks.

Also, have you run the tests to validate that this doesn't introduce regressions?

Thanks.

helmut-steiner:
Hm, looks like I just ran the streaming tests previously. The **CreateFromUri_Options** test throws an "Invalid FFmpeg option(s)" when running. All other tests pass. Big apologies for that!! I needed a quick fix for my app and as written incorporated the fixes from user dingpong. 
Gotta look into this...

Thanks for your response and the FFmpegInterop project in general!!

helmut-steiner:
So I had a bit time and tracked down the error to the line
`if (avformat_open_input(&avFormatCtx, charStr, NULL, &avDict) < 0)` in
`FFmpegInteropMSS::CreateMediaStreamSource(String^ uri, bool forceAudioDecode, bool forceVideoDecode, PropertySet^ ffmpegOptions)` which fails to open the rtsp stream. Any ideas?

khouzam:
Thanks a lot @helmut-steiner.

I'll validate, format the code and merge.

khouzam:
@helmut-steiner,

I took a look at the code and there is one aspect that I'll want to address before merging this in.

The documentation for  `avcodec_send_packet` mentions the following:
`                  Unlike with older APIs, the packet is always fully consumed,
                  and if it contains multiple frames (e.g. some audio codecs),
                  will require you to call avcodec_receive_frame() multiple
                  times afterwards before you can send a new packet.
`

Therefore, I'd like to adapt the code to make sure to call `avcodec_receive_frame` to drain the decoded frames before the next call to `avcodec_send_packet`

helmut-steiner:
Yes sure, please adapt it! Sorry for missing that!

khouzam:
I've created a new branch, helmut-steiner-ffmpeg3.1, which has @helmut-steiner changes as well as the changes in logic for the new `avcodec_send_packet` and `avcodec_receive_frame` logic. I've tried it on a few files and things look good. According to the samples that we pull out of the file and the ones that we pull out of the decoder, everything matches up.

I need to do some more testing with a bigger variety of files, but I'd like to see if other can try it and see if they find any issues.

Thanks

helmut-steiner:
Thank you!

PR #108, opened 2016-12-10
Add releasing of the stream in destructor

Romasz:
There seems to be a memory leak once you don't release the stream in destructor. This should fix that.


Romasz:
@khouzam Right, added check-up.

Issue #142, opened 2017-06-06
Memory leak in ffmpeg AVUTIL-55


mcosmin222:
I found a memory leak which traces back to AVUTIL-55.
Can anyone confirm this?

![avutil leak](https://user-images.githubusercontent.com/9284164/26850122-2079c99a-4b0f-11e7-8189-81d3ea7db1ac.PNG)

khouzam:
Hi @mcosmin222

mcosmin222:
Yes, this is simply by playing back things (with force audio decode). I managed to reproduce it with the FFMPEG.interop nuget package as well.

You can try the sample bellow. Start it and watch in task manager as memory slowly builds up (roughly 0.1 MB every 3 seconds).

[MediaPlayerCS.zip](https://github.com/Microsoft/FFmpegInterop/files/1057009/MediaPlayerCS.zip)

mcosmin222:
The problem seems to be somewhere along the path of decoding audio files. There is no leak if we don't force decoding.

khouzam:
Ok, I found the leak, we were leaking every single sample because I was using `av_freep` instead of `av_sample_free` to free them. The profiler in VS did find the samples and showed that there were about 1800 more allocations in a 30 second period and most of them were from `av_sample_alloc`.

khouzam:
@mcosmin222 can you try the fix and confirm that the leak is gone in your case?

mcosmin222:
Hmm I think we might leak packages as well.
After merging your changes, I get the following leak:

![capture 22](https://user-images.githubusercontent.com/9284164/26944682-0d6fb272-4c92-11e7-87e1-438de93758a2.PNG)

Can you check and see if you have the same problem? Maybe I didn't merge properly.

mcosmin222:
Ok ignore that. Ive run a stress test for a few hours and app remained stable. All seems good, if I can find something else I will post here.

khouzam:
After confirmation I've merged the fix.

PR #143, opened 2017-06-08
Fixing memory leak in uncompressed audio and video processing.

khouzam:
Frames allocated with av_frame_alloc need to be freed with av_frame_free

Thanks to @mcosmin222 for finding the issue. Issue #142


mcosmin222:
Looks good so far. I don't think we still need to unref the frames first, av_frame_free seems to do that automatically according to docs.

mcosmin222:
Ok, I've run the changes over night on a WP8.1 app on a phone with 1GB of memory to test long running performance and the app remained stable. I think it is good now.

PR #153, opened 2017-07-18
Dcarpente/small samples fix

dcarpente:
For some files, FFMPEG outputs audio samples that are as short as 2 or 3 ms, which was causing performance issues in our scenarios. This pull request changes MediaSampleProvider::GetNextSample to take in a minSampleDuration argument, and changes FFmpegInteropMSS to request audio samples to have duration >= 50 ms.

### DIFF
--- a/FFmpegInterop/Source/FFmpegInteropMSS.cpp
+++ b/FFmpegInterop/Source/FFmpegInteropMSS.cpp
@@ -40,6 +40,9 @@ using namespace Windows::Media::MediaProperties;
 // Size of the buffer when reading a stream
 const int FILESTREAMBUFFERSZ = 16384;
 
+// Minimum duration for audio samples (50 ms)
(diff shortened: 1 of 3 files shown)


khouzam:
Hi @dcarpente,

I've pushed a small change, the start offset was missing from your merge of master.

But also, I've tried the following and I'm seeing a huge delay now.

//const TimeSpan MINAUDIOSAMPLEDURATION = { 500000 };
const TimeSpan MINAUDIOSAMPLEDURATION = { 250000000 };

I would assume that the audio sample timestamp represents the start of the audio sample. So if I pass in a much bigger sample, such as 2 compressed samples, I would expect that they both play with the timestamp still matching. But this instead I'm seeing a huge AV-Sync issue about the size of a compressed sample. Is that normal or am I missing something?

khouzam:
After a clean build, the sync issues don't show up anymore. I think we're good.

PR #154, opened 2017-07-18
dcarpente/stream_read_buffer_fix

dcarpente:
Our usage case uses an IStream implementation that is marshalled across apartments. Because of the way buffers are copied, IStream::Read rewrites the input buffer with null characters even if no bytes are read.

FFMPEG (or at least the OGG container parser) expects FileStreamRead to leave the input ptr unmodified when bytesRead == 0. Since behavior on the buffer for IStream::Read is not defined when bytesRead == 0, we do an additional copy in a temporary buffer to guarantee the behavior FFMPEG expects.


dcarpente:
We made a change to how IMFByteStream initializes the buffer cross process in OS code, so we don't need this change any more. We should still cherry-pick part of the 531169d change in another pull request.

khouzam:
Thanks. Cherry-picked to master

PR #155, opened 2017-07-19
Updating FFmpeg submodule to n3.3.2 6d7192bcb7bb

khouzam:
I would like to update the base version of FFmpeg to the current release, but before I do this, I would like to have people try it to make sure that existing scenarios aren't broken.

Thanks.


mcosmin222:
I will update to latest ffmpeg in the weekend. I will then loop through all my test files.

But in my experience (i do this every 3 months or so), ffmpeg releases tend to be stable.

mcosmin222:
Looks good.

Issue #161, opened 2017-07-29
a bug and a question.


cansado2930:
Hello,
I have compiled the last commit of master today and If i play a mp4 video, it is played with 2x or 3x of normal speed. It is played very fast. With mp3 file, it Works correctly. It has been tested with the sample app.
And about question, I'm trying to play rtp over Multicast and it doesn't work. I tested it with ffplay and works so, What should I do in order to play rtp video from Multicast.?
Thanks you

touseefbsb:
I am having the same problem with video files it plays very fast, even the sample app is giving this problem

mcosmin222:
Can you check if the sampling rate detected by the FFMPEG interop is 2x-3x bigger than what other players report?

touseefbsb:
how can we check the sampling rate?

mcosmin222:
In ffmpeg interop you can check the values used to create AudioEncodingProfile or VideoEncodingProfile.

In windows explorer, you can usually see these values in the details tab of a files property window.

touseefbsb:
which file should we check? please clarify as you said we should check sampling rate from ffmpeginterop and also from other players, how should we do that? Thanks

touseefbsb:
I normally checked audio sample rate by right clicking video file in file explorer and going to details tab and value is "24 kHz"

cansado2930:
I have checked values of size, códec, frame rate, channels,... from audio and video and they are same from properties of file.

khouzam:
Do you have a sample that we can look at?

cansado2930:
One of videos that I have test it is this: https://1drv.ms/v/s!Av2l1Q_MdsTVggd2TyPtrQIuRMpG
I have told about 0.5 in order to know the speed of play video. But you can't use 0.5 as normal because the sound is x2x0.5 so it doesn't sound very well.

khouzam:
Thanks. I can repro the issue with the most recent code, but not with an older build so its a recent regression. Let me investigate a little bit more.

khouzam:
I've identified the code change that is causing this. In order to improve performance with small audio samples, we've tried to create bigger samples when they are too small, but it seems that for this file, the concatenation is not fully working.

For the time being, you can change the following line in FFmpegInteropMSS.cpp from 500000 to 0 and this will restore the previous logic:
```
// Minimum duration for audio samples (50 ms)
const TimeSpan MINAUDIOSAMPLEDURATION = { 500000 };
```

[...]

PR #163, opened 2017-08-10
Dcarpente/small samples fix compressed

dcarpente:
This change changes GetNextSample to only enforce minimum duration for uncompressed samples. Combining multiple packets for compressed samples can result in loss of data. This should fix issue #161.

### DIFF
--- a/FFmpegInterop/Source/UncompressedAudioSampleProvider.cpp
+++ b/FFmpegInterop/Source/UncompressedAudioSampleProvider.cpp
@@ -22,6 +22,9 @@
 
 using namespace FFmpegInterop;
 
+// Minimum duration for uncompressed audio samples (50 ms)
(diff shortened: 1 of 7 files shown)


FFmpegInterop/Source/UncompressedAudioSampleProvider.cpp
```diff
@@ -102,4 +105,42 @@ HRESULT UncompressedAudioSampleProvider::ProcessDecodedFrame(DataWriter^ dataWri
...
+
+	LONGLONG finalPts = -1;
+	LONGLONG finalDur = 0;
+
+	do
+	{
+		LONGLONG pts = 0;
+		LONGLONG dur = 0;
+
+		hr = GetNextPacket(dataWriter, pts, dur);
```
dcarpente: Thank you for the catch, fixed

khouzam:
Thanks @dcarpente

Issue #167, opened 2017-08-22
Sample duration issues


brabebhin:
I encountered a file which ffmpeg reports to have  packages of 0 length (which is wrong since we read buffer from them). I suspect it is so because ffmpeg improperly reports a block align of 0 for that file, when it is in fact 4. I think we should default to doing the mathematical calculation of buffer lengths manually when this particular condition occurs to prevent any possible wide range of bugs in ffmpeg.

I've done it for uncompressed audio sample provider.

The mathematical formulae are

block_align = num_chanels * (bit_rate / 8); bit_rate is always 16 in our case, hence it becomes num_chanel*2;

average_bytes_per_second = sampe_rate * block_align

buffer_duration = buffer_.length/ average_bytes_per_second  * 10000000;

khouzam:
Hi @mcosmin222 

Has this been addressed with #178? Can we close the issue?

mcosmin222:
hi @khouzam 
#178  never got merged. I closed it. You can cherry pick the specific commit which fixes this issue.

PR #168, opened 2017-08-22
Fixed sample duration strange cases

mcosmin222:
In some cases, ffmpeg report wrong sample durations for packages. This helps deal with that situation


mcosmin222:
This should fix #167 partially.

mcosmin222:
I will close this since it is part of #178

PR #171, opened 2017-09-18
Fixing playback of raw streams.

khouzam:
A rawstream will not have timestamps in samples, so keep a running value based on the duration of samples.


hieuvn:
Mr @khouzam, please update it on nuget, thank a ton.

Issue #175, opened 2017-10-06
Streams are stopped as soon as a single corrupted package occurs


lukasf:
Especially with broadcast streams, broken packets can occur, but also normal files can have a broken packet here and there (bad encoder, whatever). With the current code, if a packet cannot be decoded, a "null" value is returned as Sample. If MediaStreamSource gets a "null" Sample on a stream, this is always considered end-of-stream, so playback of that stream is stopped for good and will not recover.

I would propose to add an option "BrokenPacketThreshold". If you assign e.g. 10, then this means that up to 10 broken packets can occur in a row before the stream is stopped. Default would be 0 (current behavior), -1 would mean skip all packets until a good one is found.

I have already tested a quick hack of this with partly corrupted mpeg files, works well for me.

mcosmin222:
I have a fix for this. PR latter on.

lukasf:
Cool, looking forward to that.

mcosmin222:
But it is pretty much the same implementation as what you said: it tries to skip over 10 broken packages in a row. if more than 10 then aborts with null sample. So if you already have something, feel free to upload.

lukasf:
There are some more things to consider here: If we return a null sample (because more than X broken packets have been read), we must disable the stream internally. Otherwise, the stream's packet queue will be flooded. Packets are still enqueued by the reader but no samples are read out anymore by MSS so queue grows infinitely.

I will come up with a PR for this soon fixing both problems.

lukasf:
Please check my solution #182.

Issue #174, opened 2017-10-06
CreateFFmpegInteropMSS methods will get out of hand if we add a lot more options


lukasf:
There are already a whole bunch of parameters already in these methods. I have some new features in mind which would require to add even more parameters. If we continue to evolve the library and add all new options here, we will run into problems, this will be confusing.

I would propose to move to a different initialization model, like this:

```
var ffmpeg = new FFmpegInteropMSS();
ffmpeg.ForceVideoDecode = false;
ffmpeg.ForceAudioDecode = true;
ffmpeg.AllowMultipleAudioStreams = true;
ffmpeg.CorruptPackageThreshold = 100;
await ffmpeg.OpenStreamAsync(stream);  // this could be async to prevent UI blocking
var mss = ffmpeg.GetMediaStreamSource();
```

This is more like what you expect from framework WinRT classes.
The old CreateFFmpegInteropMSS could be left for compatibility reasons.

What do you think?

khouzam:
That sounds very reasonable. I'd love to start the discussion about this. I do want this project to move in the direction that the community wants.

mcosmin222:
Maybe we should have a builder instead of exposing them directly to the MSS

lukasf:
I see FFmpegInteropMSS as the builder, and you get the result with GetMediaStreamSource() (which is what you are really interested in).

lukasf:
Actually, I'd rename GetMediaStreamSource() to CreateMediaStreamSource(), so after OpenAsync() you could inspect and configure the streams (in case you have multiple audio streams, you could e.g. set force decode per-stream), and then finally CreateMediaStreamSource().

mcosmin222:
Not really, the FFMpegInteropMSS also has some stuff needed by the MSS to work properly (like handling its events). So it is sort of like a builder but more than a builder.

lukasf:
True, it is both builder and stream handler. But opening a file and handling MSS events is very closely coupled, I am not sure if this can be separated in a good way. Or if you see the builder as just collecting property values and passing them to FFMpegInteropMSS constructor then I do not see so much improvement in that.

mcosmin222:
The problem with having public setters for properties is that the ffmpeginterop object is immutable in nature. You can't just set ForceAudioDecode mid way during playback. Doing that would force you to recreate the entire media stream source, and in order for it to be transparent for consumers, you would also have to seek to its last known position. 

[...]

PR #178, opened 2017-10-07
Skip corrupted packages

mcosmin222:
Code to skip up to 10 corrupted packages in a row.  Fix for #175 
This can be extended to have a public setter property on the ffmpeg interop MSS if we decide to do so, see #174 

This has been in production for my app for almost 3 weeks with no known side effects. 
Also helps with files with corrupted headers resulting into 0 block aligns.

### DIFF
--- a/FFmpegInterop/Source/UncompressedAudioSampleProvider.cpp
+++ b/FFmpegInterop/Source/UncompressedAudioSampleProvider.cpp
@@ -45,8 +45,8 @@ HRESULT UncompressedAudioSampleProvider::AllocateResources()
 		int64 outChannelLayout = av_get_default_channel_layout(m_pAvCodecCtx->channels);
 
 		// some files report sample duration 0. They also happen to have the m_pAvCodecCtx block_align 0
-		blockAlign = m_pAvCodecCtx->channels * (16 / 8);
-		avgBytesPerSec = m_pAvCodecCtx->sample_rate * blockAlign;
+		m_blockAlign = m_pAvCodecCtx->channels * (16 / 8);
+		m_avgBytesPerSec = m_pAvCodecCtx->sample_rate * blockAlign;
 		
 		// Set up resampler to convert any PCM format (e.g. AV_SAMPLE_FMT_FLTP) to AV_SAMPLE_FMT_S16 PCM format that is expected by Media Element.
 		// Additional logic can be added to avoid resampling PCM data that is already in AV_SAMPLE_FMT_S16_PCM.
--- a/FFmpegInterop/Source/UncompressedAudioSampleProvider.cpp
+++ b/FFmpegInterop/Source/UncompressedAudioSampleProvider.cpp
@@ -46,7 +46,7 @@ HRESULT UncompressedAudioSampleProvider::AllocateResources()
 
 		// some files report sample duration 0. They also happen to have the m_pAvCodecCtx block_align 0
 		m_blockAlign = m_pAvCodecCtx->channels * (16 / 8);
-		m_avgBytesPerSec = m_pAvCodecCtx->sample_rate * blockAlign;
+		m_avgBytesPerSec = m_pAvCodecCtx->sample_rate * m_blockAlign;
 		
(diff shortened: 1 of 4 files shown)


lukasf:
Please check #182 for my solution.

mcosmin222:
I am not sure how the ffmpeg reader will continue reading. if the threshold exceeds, it will send a null sample and will trigger media ended event, which should dispose the media stream source. It is not any different than how it is right now. You get 1 bad package, you send null sample. All I do is to send null sample after 10 packages.

lukasf:
No, if you send a null sample, MediaStreamSource will only stop requesting samples from that specific stream. But if there are other streams which continue to deliver samples, MSS will continue playing those other streams. And in that case, the queue of the stopped stream will be flooded. You are right in saying that the same problem exists without your PR. You did not introduce this bug, it was there before. But I say we should fix both issues, any my PR does that.

I have a file where only the audio stream has some broken packets every now and then. With the current solution, audio stream is stopped but video keeps playing and audio queue grows and grows without end. We cannot risk out-of-memory exceptions just because of a bad file.

lukasf:
And additionally, your solution only covers the uncompressed streams. If broken packets occur on a compressed stream (e.g. force audio decode = false), they are not skipped with your solution. My solution works on all streams because it is implemented on a different layer.

mcosmin222:
Well, if you found one of these weird files, feel free to upload it and have it as a test (we need a stock of weird files to test edge cases). I never encountered such a situation so far, but if you did, and your solution covers more situations, then I am all for it.

lukasf:
The file is way too big to put it in the repo. Any file with more than one stream and at least one broken frame will show this behavior. You can try yourself with one of the files you have (you had the problem too with some files, right?). Just set your threshold to 0, wait for the broken packet to occur, then put a breakpoint in FFmpegReader.ReadPacket and see how it will fill up the queue of that stream. FFmpegReader does not know that the stream is not processed by MSS anymore and continues to call QueuePacket() on the stream. Only when all streams have stopped, MSS stops playback. If the file is short enough, it is not a problem. But on a long file, it could lead to out of memory situations.

mcosmin222:
Hmm strange. Are you sure about that?

By eye balling the code, the ffmpeg reader should not read a package unless it is asked to and if the mss stops asking for packages in a stream, the reader shouldnt read either. I can't check it myself right now.

I do agree that the unused stream should be flushed.

lukasf:
Audio stream stops due to broken packet. But MSS continues to ask video stream for samples. So video stream calls FFmpegReader.ReadPacket, which will then read a packet and put it either in video or audio stream's queue, depending on the packet's stream id. Please check implementation of ReadPacket, it does not read a specific stream packet, it just reads the "next" packet, checks which stream it is for, and puts it into the respective stream's queue (without knowing if a stream is still active or not)...

mcosmin222:
Ok, if it is that way, and reads the stream anyway,maybe we can find a way to skip unknown numbers of broken frames and "reactivate" the broken stream when it is functional again?

I have been trying to get rid of that hard coded value for some time now.

mcosmin222:
Maybe provide a mock frame instead of a null?  IIRC, there are ways to tell apart EOF from actual bad data.

mcosmin222:
Anyway, i will fix my ffmpeg reader latter this week. Thanks for spotting that.^^

lukasf:
Why don't we just take my PR then? It is already finished, supports both compressed and uncompressed streams, stops stream if threshold is hit to prevent memory increase. Most of the skipping is done in GetNextPacket, which makes it work automatically with both compressed and uncompressed streams. Additionally, I do take care that UncompressedAudioSampleProvider does not further concatenate samples once a broken packet occured (your current solution leads to additional distortions because it concatenates frames that are not ment to be played in that order - instead, we need to create a new sample after the gap with correct new time stamp). Plus, I also set the Discontinuous flag on each sample if broken packets have occured. This helps downstream decoders, so they know they have to reset their decoder state (only required for compressed streams).

I mean, sure you can re-implement this all again. But just saying, it's all been done already. You can download the files form my PR and take a look.

About those mock frames: Maybe it's a good idea, it would help recover after long errors. But it only works for uncompressed streams. You could implement mock frames for uncompressed streams, on top of my PR, but my solution (or a similar one) would still need to be in place for compressed streams.

PR #182, opened 2017-10-10
Skip broken packages - different approach

lukasf:
I do the skipping of broken packets at a higher level in GetNextPacket. This makes it work automatically for all sample providers. So there is no need to implement it explicitly in each provider (which can be easily forgotten when new providers are added).

Second, once the threshold has been reached, the stream will be flushed and stopped (m_isEnabled = false). This prevents FFmpegReader from pushing new packets into the queue.

### DIFF
--- a/FFmpegInterop/Source/MediaSampleProvider.cpp
+++ b/FFmpegInterop/Source/MediaSampleProvider.cpp
@@ -34,6 +34,7 @@ MediaSampleProvider::MediaSampleProvider(
 	, m_startOffset(AV_NOPTS_VALUE)
 	, m_nextFramePts(0)
 	, m_isEnabled(true)
+	, m_isDiscontinuous(false)
 {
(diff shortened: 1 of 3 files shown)


lukasf:
Updated: I realized that at least the UncompressedAudioSampeProvider must be adapted as well, because it concatenates packets and we do not want to concatenate packets that were not read in order. Updated PR.

mcosmin222:
Finally got around to test it. Looks good so far.

lukasf:
@mcosmin222 Have you been able to test the PR? I would like to do another PR...

mcosmin222:
Yes, it works OK, no obvious flaws so far.

lukasf:
Can we then decide to take this PR and close the other?

lukasf:
I have a bunch of other things I would like to push into the repo...

mcosmin222:
@khouzam this PR is safe to merge.

Issue #172, opened 2017-09-26
[Bug]Sample presentantion timestamp 


brabebhin:
Hello,

I discovered a rather disturbing bug. Assume we have a standard ffmpeg interop MSS, and a MediaPlayer object as consumer.
If the MediaPlayer object has AutoPlay false, and you try to seek programmatically before the first sample is delivered, the subsequent samples have the wrong presentation time.

mcosmin222:
The problem is located in MediaSampleProvider::GetNextPackage


	if (SUCCEEDED(hr))
	{
		// Write the packet out
		hr = WriteAVPacketToStream(writer, &avPacket);

		if (m_startOffset == -1)
		{
			//if we havent set m_startOffset already
			DebugMessage(L"Saving m_startOffset\n");

			//in some real-time streams framePts is less than 0 so we need to make sure m_startOffset is never negative
			m_startOffset = framePts < 0 ? 0 : framePts;
		}

		pts = LONGLONG(av_q2d(m_pAvFormatCtx->streams[m_streamIndex]->time_base) * 10000000 * (framePts - m_startOffset));

		dur = LONGLONG(av_q2d(m_pAvFormatCtx->streams[m_streamIndex]->time_base) * 10000000 * frameDuration);
	}

m_startOffset will always cause samples to be presented by the beginning of the stream instead of their actual presentation time. This makes seeking before playing or MediaPlaybackItems with a start position other than 0 unusable.
I am not sure what the purpose of m_startOffset is supposed to be.

mcosmin222:
Changing to this fixes my use case

```
		if (m_startOffset == -1)
		{
			//if we havent set m_startOffset already
			DebugMessage(L"Saving m_startOffset\n");

			//in some real-time streams framePts is less than 0 so we need to make sure m_startOffset is never negative
			m_startOffset = m_pAvFormatCtx->streams[m_streamIndex]->start_time < 0 ? 0 : m_pAvFormatCtx->streams[m_streamIndex]->start_time;
		}
```

mcosmin222:
```
		if (m_startOffset == -1)
		{
			//if we havent set m_startOffset already
			DebugMessage(L"Saving m_startOffset\n");

			//in some real-time streams framePts is less than 0 so we need to make sure m_startOffset is never negative
			m_startOffset = m_pAvFormatCtx->streams[m_streamIndex]->start_time < 0 ? 0 : m_pAvFormatCtx->streams[m_streamIndex]->start_time;
[...]

PR #195, opened 2017-12-08
Sample Presentaion Timestamp

khouzam:
Fix to use the start_time of the stream for the startOffset instead of the first packet timestamp. From issue #172

### DIFF
--- a/FFmpegInterop/Source/MediaSampleProvider.cpp
+++ b/FFmpegInterop/Source/MediaSampleProvider.cpp
@@ -42,7 +42,7 @@ MediaSampleProvider::MediaSampleProvider(
 HRESULT MediaSampleProvider::AllocateResources()
 {
 	DebugMessage(L"AllocateResources\n");
-	m_startOffset = avFormatCtx->start_time == AV_NOPTS_VALUE ? 0 : LONGLONG(avFormatCtx->start_time * 10000000 / double(AV_TIME_BASE));
+	m_startOffset = m_pAvFormatCtx->start_time == AV_NOPTS_VALUE ? 0 : LONGLONG(m_pAvFormatCtx->start_time * 10000000 / double(AV_TIME_BASE));
 	m_nextFramePts = 0;
 	return S_OK;
 }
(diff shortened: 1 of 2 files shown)


mcosmin222:
Looks OK so far, I will give it some more tests tomorrow.

mcosmin222:
Good to merge.

FFmpegInterop/Source/MediaSampleProvider.cpp
```diff
@@ -223,17 +223,17 @@ HRESULT FFmpegInterop::MediaSampleProvider::GetNextPacket(DataWriter ^ writer, L
 			DebugMessage(L"Saving m_startOffset\n");
 
 			//in some real-time streams framePts is less than 0 so we need to make sure m_startOffset is never negative
-			m_startOffset = framePts < 0 ? 0 : framePts;
+			m_startOffset = m_pAvFormatCtx->streams[m_streamIndex]->start_time < 0 ? 0 : m_pAvFormatCtx->streams[m_streamIndex]->start_time;
```
lukasf: How about using (positive) framePts as fallback if we do not have start_time? Or can we be sure that start_time is there?

reego-fr: Hi,

not sure if it can help but https://github.com/Microsoft/FFmpegInterop/commit/fd353fe03bfe40cd3fdae8df86b644d5b6be083f was tested a lot and proved to be working. 
I don't know what made @khouzam choose another implementation and I won't link to the original thread because it became a real mess because of other non related issues !

mcosmin222: I had another way of fixing this. But this PR works. the only thing I would change is to move the resolving of start time in the AllocateResources() method

reego-fr: I don't know why only positive offsets were taken into account in the first place. Doesn't it make sense to apply the start offset whatever the value different of AV_NOPTS_VALUE is ? 
I never encountered the case myself but I think start_time would be AV_NOPTS_VALUE if it could not be set by ffmpeg.

reego-fr:
It is good to use ffmpeg start_time value to fix the issue reported by @mcosmin222 but there are still minor fixes to correctly play and be able to seek to correct timestamps on MPEG-PS VOB files and MPEG-TS stream dumps. 
Should I create a new pull request with these changes or push changes here ?

FFmpegInterop/Source/MediaSampleProvider.cpp
reego-fr: FYI, I encountered negative start_time value in a MPEG-TS stream dump that contains a PTS rollover.

mcosmin222:
Yes, please do so, since you have better kbowledge about it. Thanks

reego-fr:
I merged the older fix, I'm not sure why the username is wrong but the code should be good.

mcosmin222:
You fix is similar to what i had, but didnt include all the cases. It works. Good to merge.

reego-fr:
I changed av_seek_frame to use the stream's start_time which makes more sense and hopefully may solve issues occuring on playback start with some files such as one found here https://github.com/Microsoft/FFmpegInterop/issues/201

mcosmin222:
@reego-fr , is this working OK for you now? My use case is fixed, so perhaps @khouzam  can merge it is everything is OK on your end.

PR #198, opened 2017-12-13
Fixed sample duration strange cases

brabebhin:
In some cases, ffmpeg report wrong sample durations for packages. This helps deal with that situation

### DIFF
--- a/FFmpegInterop/Source/UncompressedAudioSampleProvider.cpp
+++ b/FFmpegInterop/Source/UncompressedAudioSampleProvider.cpp
@@ -45,7 +45,7 @@ HRESULT UncompressedAudioSampleProvider::AllocateResources()
 		int64 outChannelLayout = av_get_default_channel_layout(m_pAvCodecCtx->channels);
 
 		// some files report sample duration 0. They also happen to have the m_pAvCodecCtx block_align 0
-		blockAlign = m_pAvCodecCtx->channels * (16 / 8);
+		blockAlign = m_pAvCodecCtx->channels * 2;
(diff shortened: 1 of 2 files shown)


lukasf:
Looks fine.

mcosmin222:
@khouzam , to which case are you referring exactly? The calculation of block_align?

FFmpegInterop/Source/UncompressedAudioSampleProvider.cpp
```diff
@@ -134,7 +138,14 @@ MediaStreamSample^ UncompressedAudioSampleProvider::GetNextSample()
 			{
 				finalPts = pts;
 			}
-			finalDur += dur;
+			if (dur == 0 && dataWriter->UnstoredBufferLength != 0)
+			{
+				finalDur = (dataWriter->UnstoredBufferLength / avgBytesPerSec) * 10000000;
```
lukasf: I think @khouzam means this line. Maybe better do it like this?

`finalDur = ((long long)dataWriter->UnstoredBufferLength * 10000000) / avgBytesPerSec`

I am not sure if the long long cast is neccessary, but we need to be sure that we don't get overflow.

mcosmin222: I will change all of them this weekend. I haven't had the time to do much testing and development on this lately :(

lukasf:
Somehow, your last changeset changed the whole file. Did you perhaps accidentially change line endings? Not such a big deal for me, but it seems to break GitHub diff tool.

Anyways, changes look fine to me.

mcosmin222:
Strange. I did edit it in the online editor, maybe that is what caused the issue.

PR #199, opened 2017-12-13
Add HEVC passthrough for hardware acceleration

lukasf:


lukasf:
I found some issues with files from broadcast streams. Will update the PR once this is solved.

lukasf:
Okay this works now with all test files I have. But I somehow messed up my commit history. Don't know what went wrong. The result set is still correct. Do you use squash commits? Otherwise I can make a new PR with clean changeset.

lukasf:
Ah well I just created a new, clean PR... Closing this one.

PR #200, opened 2017-12-13
Add HEVC passthrough for hardware acceleration - clean

lukasf:
Please note that I had to upgrade UWP target platform to 10.0.15063 (creators update).

### DIFF
--- a/FFmpegInterop/Source/FFmpegInteropMSS.cpp
+++ b/FFmpegInterop/Source/FFmpegInteropMSS.cpp
@@ -21,6 +21,7 @@
 #include "MediaSampleProvider.h"
 #include "H264AVCSampleProvider.h"
 #include "H264SampleProvider.h"
+#include "HEVCSampleProvider.h"
 #include "UncompressedAudioSampleProvider.h"
 #include "UncompressedVideoSampleProvider.h"
 #include "CritSec.h"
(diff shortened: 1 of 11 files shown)


lukasf:
@khouzam Since the actual parsing of NAL and SPS/PPS is different for each of the sample providers, I do not see so much reuse from a common base class.

But there is one thing that should perhaps be aligned with all sample providers: As per my understanding, the extradata is part of the file header, and as such it is valid for the whole file. That would mean that it only needs to be sent to the decoder once, at the beginning of playback. But the current implementation for H264 parses and sends the extradata again and again with each key frame. Is this done on purpose? I have tried to change H264 decoders to only send extradata once, and it works well with all my files, also seeking and all works well. My current HEVC implementation also only sends extradata once. If there is a reason why extradata must be repeated (maybe it can change during playback in certain scenarios?), the I should perhaps do that for HEVC as well. If there is no reason, I could change the H264 implementation to also only send this once, to prevent unneccessary parsing and data throughput. Please let me know what you think.

reego-fr:
I read SPS and PPS may change within a file, some encoders may be leveraging this capability. It must be possible to skip pushing them to the decoder if no parameters changed but you would still have to parse them to see if they changed since the beginning of the played range.

lukasf:
The AVCodecContext->extradata should not change, it is what is found in the file header. I was under the impression that extradata changes during playback will be sent through normal NAL packets. But after consulting google again and checking ffmpeg sources, I see that for both h264 and hevc, extradata changes can come as "side_data" on the AVPacket. So now it looks like, instead of repeatedly parsing AVCodecContext->extradata during playback, we need to check for AVPacket->side_data for each packet.

You can also look at hevcdec.c and h264dec.c. Both parse AVCodecContext->extradata only once during init (playback start), and then check packet->side_data on each packet.

Also after checking ffmpeg sources, I think the h264 parser does not correctly parse extradata. It assumes only one SPS and PPS packet is contained, and it uses fixed nal_length_size. But ffmpeg sources also check for multiple packets, and they retrieve nal_length_size from extradata (both like I do in hevc).

I will try to add support for side_data in HEVC. H264 can be fixed in a separate PR.

lukasf:
Now I see enough potential for a common base class. Latest commit makes a lot of reuse between H264 and HEVC. It also adds correct handling of extradata changes through side_data for both HEVC and H264, plus improved H264 extradata and NAL packet parsing.

reego-fr:
It looks good to me, I don't know a lot about h264 and hevc but tested it and it worked.

Issue #202, opened 2017-12-28
FFmpegInterop really struggles to play some 1080p videos and all 4k videos


viswanathr88:
iPhones now captures videos in 1080p and 4k and as a result they have become so common. I have my own uwp video project where I'm trying to play such videos. I had lot of issues with framerate, that's why I checked this project. 
1. First of all, audio and video pipeline are not split up into separate threads in ffmpeg interop. That way, cores can be better utilized. Also, decoding and scaling the frame too. This made 1080p videos play much smoother, but 4k was still pretty poor.
2. Then I able to solve most of the frame rate issues by asking FFMPEG to multi-thread the decode process (which was also recently added by FFmpegInterop). But the problem now seems to be sws_scale. It takes ~30ms on my machine (4 core 8 thread - i7) to scale one frame.
Note: I'm only talking about the case where Ffmpeg does the decoding and not Windows.

My question is: Is there any way by which sws_scale can be avoided all together and somehow we could render the YUV frame onto the screen directly? I feel that's the only way we could keep up the frame rate. VLC Store app is somehow able to do a fantastic job, even using less CPU.

mcosmin222:
Have you tried the latest verion of this repository? We have multithreaded video decoding now.

About VLC, my guess is that they use hardware acceleration for that, which this library can also do.

viswanathr88:
I posted this only after trying it out with the latest version of ffmpeginterop.
1. Pick a 4k video recorded in an iPhone or android phone
2. Start ffmpeginterop (c#) and set Force Decode Video and Force Decode Audio to be true.
3. Play the 4k video.

Ffmpeg can play the video for 2 seconds and then buffers for the next 4 seconds, then plays 2 seconds and buffers again and so on.

mcosmin222:
The only way to render the YUV frame directly is to use directx. This is sadly not the purpose of this library so far, we are just tieing things into a consumer MediaPlayer(Element).

VLC does something similar with Video decoding. But VLC on the other hand has its own custom media players.

reego-fr:
Hi, I am surprised sws_scale can be so slow on a desktop but if the following still applies it could be replaced by a faster implementation if it is possible for you to go GPL. [This link](http://www.ffmpeg-archive.org/sws-scale-slow-with-big-resolutions-td2272531.html) contains a suggestion to compile ffmpeg with --enable-gpl. 
The best of course would be to get rid of it when possible, but I am not sure what pixel formats the current platforms are limited to.

[...]

PR #204, opened 2018-01-01
Vastly optimize video decoding performance of high-res (4K) files

lukasf:
- Eliminate sws_scaler where possible (MF will do scaling on GPU)
- Use Platform::ArrayReference to avoid unneccessary copy
- Copy sws_scaler output in one operation (contiguous buffer)


lukasf:
See discussion in #202. With this patch, I am able to software decode 4K HEVC files 100% smooth, which were previously absolutely unplayable. Even 4K Hi10P files play well. Performance for HD videos has also improved (lower CPU usage).

mcosmin222:
Looks good.

reego-fr:
Looks good but apparently yuvj420 and yuv420 store levels differently. Adding videoProperties->Properties->Insert(MF_MT_VIDEO_NOMINAL_RANGE, (uint32)MFNominalRange_0_255); prevents clipping in the highlights and shadows when the decoder output is yuvj420. 
Another thing non related to these changes but I noticed while testing a yuvj420 file is that the colors seem a bit different when using MF video decoding (h264) compared to VLC. Did you notice anything similar ?

lukasf:
Thanks for the hint with the levels! I originally compared my output with MPC-HC and found no difference in levels so I thought they were correct. But it looks like MPC-HC also does not treat levels correctly. Only when comparing with other video apps now, I saw that levels are really off. I updated the PR as you said, this fixes the levels. The old solutio with sws_scaler also had the levels wrong, looks like it does not update the levels when changing pix format.

And yes I also noticed already that colors are slightly off for HEVC files! Can h264 files have YUVJ as well? I only saw that pix format on HEVC files. To me it looks like green is slightly oversaturated while red is slightly undersaturated. Not sure where this comes from. Tried playing with MF_MT_YUV_MATRIX but it did not have any effect.

lukasf:
After checking the colors again with the levels fixed, I do not see much of a difference anymore. There is still a very slight difference, but it is even hard to tell what the difference is. I was not able to get rid of it, and did not find any help on google. For me, the difference is too small right now to further look into this, but feel free to open an issue if it is a problem for you.

reego-fr:
I will try to find more samples files to try to figure out what can be wrong, I am not even sure VLC output is what my sample should reallly look like. 
It is apparently the transfer characteristic of "gamma2.2" in my file that introduce this inconsistency but I could not see any change when playing with MF_MT_TRANSFER_FUNCTION attribute. 
I will probably open a new issue once I get more testing material.

mcosmin222:
Try playing the file in windows media player classic with ffmpeg codec pack. It should give you a good idea about the colors.

reego-fr:
I can not reproduce the blatent color shift I am seeing on another machine, it's probably not a big deal considering my weird test setup. 
Anyway, as I understand the color matrix or other parameters should influence yuv<->rgb conversion, so I think one time or another we will have to care about that in the case of uncompressed video playback.

reego-fr:
One last thing about Iuyv uncompressed support. Did you test it under Windows 8.1 Phone and desktop ? I wondered about retrocompatibility but just realized I could not build those projects anymore and test that.

mcosmin222:
At this point I don't think backwards compatibility should be an issue anymore. Perhaps there could be a branch for windows (phone) 8.1 specifically. Microsoft abandonment of these platforms makes it hard for us to use new media features in newer OS version while keeping it compatible with the older versions.

I've personally stopped updating the WP8.1 ffmpeg interop in my app.

lukasf:
When implementing this, I checked MSDN docs for platform support of these formats. The output formats should be supported on all platforms. Only limitation is RGBA output on WP8.1, where the alpha channel will be dropped during playback (but it should still render).

I agree with @mcosmin222 about backwards compatibility. I already planned to submit an issue to remove support for Windows 8 and WP 8 and update the project files to VS2017 toolkit. A branch can be created for legacy apps. I don't think that anyone here really cares about Windows 8 or even WP 8, and no one would start new apps for those platforms.

lukasf:
I just created #206 for dropping W(P)8.1 support. Please post your opinions there.

FFmpegInterop/Source/UncompressedVideoSampleProvider.cpp
```diff
@@ -27,18 +27,29 @@ extern "C"
...
 	{
-		m_rgVideoBufferLineSize[i] = 0;
-		m_rgVideoBufferData[i] = nullptr;
+	case AV_PIX_FMT_YUV420P:
+	case AV_PIX_FMT_YUVJ420P:
+		m_OutputPixelFormat = AV_PIX_FMT_YUV420P;
+		OutputMediaSubtype = MediaEncodingSubtypes::Iyuv;
+		break;
+	case AV_PIX_FMT_YUVA420P:
+		m_OutputPixelFormat = AV_PIX_FMT_BGRA;
```
reego-fr: Isn't there a typo on AV_PIX_FMT_BGRA ?

lukasf: Nope, ffmpeg BGRA = MF ARGB

reego-fr: Ok, then it is AVPixelFormat::AV_PIX_FMT_RGBA in WriteAVPacketToStream that's wrong no ?

lukasf: Fixed, thanks. It still worked (by chance) because LineSize[1] is 0 for BGRA...

mcosmin222:
@khouzam , Maybe it is time for the merge?

khouzam:
Let me find a bit of time to go over the code.

lukasf:
I accidentially pushed a wrong changeset into this branch. Reverted and merged latest master branch.

khouzam:
Thank you, I was going to say, let's get the frame grabber feature in a separate branch 😊

lukasf:
Yes, I really can't tell how it ended up here, just found out by coincidence :)

reego-fr:
Hi,
I actually found there is a small issue with direct yuv420p playback. I found files that decode to planes larger than the stream width (i.e. linesize is larger than width). 
On the media foundation side I tried setting MF_MT_DEFAULT_STRIDE attribute on the sample or on the videoStreamDescriptor but it didn't look like it was doing anything.

lukasf:
@reego-fr Can you upload a file somehere?

reego-fr:
Here is one https://1drv.ms/u/s!Av6YXBDKaOdvgutq2u-I2SIafHaTrw 
I don't know if it is limited to non square pixel aspect ratio files but it looks related.

reego-fr:
It happens on 1:1 files too. It is really only dependent on the decoded frame width.

lukasf:
Huh, well, I am also a bit lost here. I tried various ways, not only with MF_MT_DEFAULT_STRIDE but also MF_MT_MINIMUM_DISPLAY_APERTURE and MF_MT_GEOMETRIC_APERTURE. All those don't have any effect at all on the output.

@khouzam Do you think you can get someone from the media team look into that? I also created a post on the msdn forums:

https://social.msdn.microsoft.com/Forums/en-US/6247bc89-f741-4203-80d8-a0db07e3fa0d/uwp-win10-mediastreamsource-setting-mfmtdefaultstride-does-not-have-any-effect?forum=wpdevelop

lukasf:
I checked in a workaround for this problem. If the decoder wants to add stride, I now force use of sws_scale. This seems to work well so far. But of cause it is only a workaround. Best would be if we could use the stride MF attribute.
@reego-fr Can you test the latest update with your files as well?

mcosmin222:
If this is indeed a bug, it will probably be fixed in the next major windows update, so your workaround is there to stay.

lukasf:
Yes I guess so.

reego-fr:
It works for me too. Thanks.

lukasf:
I found out how to use MF_MT_MINIMUM_DISPLAY_APERTURE to work around MF_MT_DEFAULT_STRIDE not working. Latest checkin does the trick!

mcosmin222:
I think there was a misunderstanding with the default stride actually does. The documentation is a bit misleading. The default stride should be the actual size of the video frame. I this case, 720 not 768. Msdn mentions that default stride only has effect when there are no padding bytes. I think this attribute is intended for special case video frames and that your solution is the more generic one, not the other way around.

lukasf:
Ah, the MS Media guy also told me to use default stride of 768, but he also recommended setting aperture as well. That is why I tried again with the latter and got it working in the end...

reego-fr:
It works for me

lukasf:
Hey guys, I did not really think that I could get this working so quickly: Right now I already have a solution running on a different branch, which totally elimitates any buffer copying. When no scaler is used, the decoder directly decodes into what will be the target IBuffer. When scaler is used, the scaler will render into the target IBuffer. Plus, BufferPools are used for both solutions, to avoid unneccessary malloc/free, this helps keep memory consumption stable. No more memory copy at all, no DataWriter, this is as efficient as it can get.

If you have some time, please check out the branch "lukasf/experimental-direct-buffer", and let me know how it works for you. It already includes all other fixes which were done here (stride problem, color range, etc).

Not sure how we should proceed. Shall we keep this PR, or directly switch to the new, more advanced solution? It still needs some more testing, and some cleanup is needed. But for me, it works pretty well already.

mcosmin222:
I will test it asap. Meanwhile, have you tried a long running file to make sure there are no memory leaks?

lukasf:
Memory consumption is rock stable from what I see.

reego-fr:
It's really nice. No playback issues to report so far.

lukasf:
@reego-fr Cool, that's good to hear!

I have cleaned up the code and made a new branch lukasf/direct-buffer-clean. I also opened a new PR #216 so let's discuss the new solution over there. Maybe we can close this PR if we decide to jump to direct buffer approach.

PR #213, opened 2018-01-13
Fix no seek after file end bug

lukasf:
I noticed that you cannot seek anymore once a file has reached the end. Looks like I introduced this with the corrupt packet skipping. Streams would get disabled on end of file, but never enabled again. This PR will reset the "enabled" flag on seek, so all disabled streams will process data again after a seek.


reego-fr:
It fixes it for me too. Thank you for the packet skipping code btw, it proved to be working on my side too.

PR #214, opened 2018-01-16
Set default buffer time to zero

lukasf:
For normal local file playback, a buffer is not needed at all. Having the default buffer of 1 second gives a bad impression on anyone trying the lib. They will think this lib is slow, because opening files and seeking is so slow. They will see very high RAM usage (at least with high res files). Multiple issues have already been opened on these topics. Let's fix this be setting buffer time to zero by default. 

MSS buffer should only be used in very specific scenarios (e.g. streaming). In that case, the dev needs to try and find out if and which buffer size works best for his sceanrio. But most devs just want this for local file playback, and for them, zero buffer is best.

### DIFF
--- a/FFmpegInterop/Source/FFmpegInteropMSS.cpp
+++ b/FFmpegInterop/Source/FFmpegInteropMSS.cpp
@@ -490,15 +490,10 @@ HRESULT FFmpegInteropMSS::InitFFmpegContext(bool forceAudioDecode, bool forceVid
 		}
 		if (mss)
 		{
+			mss->BufferTime = { 0 };
 			if (mediaDuration.Duration > 0)
 			{
-				mss->Duration = mediaDuration;
 				mss->CanSeek = true;
-			}
-			else
-			{
-				// Set buffer time to 0 for realtime streaming to reduce latency
-				mss->BufferTime = { 0 };
 			}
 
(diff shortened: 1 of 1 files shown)


mcosmin222:
It is worth noting that setting the buffer time in media stream source does not disable buffering at MF level, just at the application level. In order to remove any traces of buffer latency, you need to set the RealTimePlayback in MediaPlayer object to true.

Regarding the buffer time for online streaming purposes, I saw some ffmpeg protocols can insert custom properties to control the buffer time internally inside ffmpeg decoders. So perhaps we can keep it at 0 regardless, and the buffer time be manipulated only internal to ffmpeg, and this can be manipulated through custom property injections.

lukasf:
And also, if buffering is really desired for a streaming scenario, it is a lot more efficient to buffer the source data stream, instead of buffering decoded samples (which unneccessarily blows up RAM usage). So even for streaming, using MSS buffer is not a good idea. Better buffer the stream (or let ffmpeg buffer it for you, as @mcosmin222 noted).

mcosmin222:
Seems BufferTime introduced a new bug. This has to be the most useless property exposed in winRT.

As a side note, for WP8. 1 users, buffer time has to stay 0, otherwise the media stream source stops working after a pause.

PR #215, opened 2018-01-22
Direct buffer decoding

lukasf:
This PR further improves video decoding performance over #204 by eliminating all unneccessary copy operations. Decoder (or scaler, if required) will directly render into output buffer. BufferPool is used to keep memory usage stable (as opposed to allocating new memory for every frame).

If a frame size change is detected which exceeds the direct decoding buffer size, the solution automatically switches to scaler. I don't have any file to test this, but I think it could happen on broadcasts or maybe also incorrectly merged files.

One more thing I noticed: When we use Iyuv output, the DXVA hardware deinterlacer does not work, making interlaced mpeg2 files look a bit worse (e.g. the simpsons file). So I added code to enforce nv12 output for mpeg2. Hope this is a good idea.

### DIFF
--- a/FFmpegInterop/Source/FFmpegInteropMSS.cpp
+++ b/FFmpegInterop/Source/FFmpegInteropMSS.cpp
@@ -26,6 +26,7 @@
 #include "CritSec.h"
 #include "shcore.h"
 #include <mfapi.h>
+#include <dshow.h>
 
 extern "C"
 {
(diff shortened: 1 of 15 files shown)


lukasf:
Ah damn, I wanted this to only have a single commit. I guess I'm still pretty bad with Git.

lukasf:
Closing this one. Please check #216

Issue #208, opened 2018-01-04
HEVC buffering a lot in phones!


ramtinak:
Hi, I have an local HEVC movie (720p). In pc there is no buffering while playing this movie, but in my phone, there's a lot of buffering!
Movie info:
```
General
Unique ID                      : 42348663308122598639254619844544870105 (0x1FDC0E4DF44EC74CDE088474C138B6D9)
Complete name                  : C:\Users\Ramti\Downloads\Video\Man_with_a_Plan_S02E06_x265_720p.mkv
Format                         : Matroska
Format version                 : Version 4 / Version 2
File size                      : 128 MiB
Duration                       : 20 min 49 s
Overall bit rate               : 858 kb/s
Movie name                     : Man with a Plan S02E06 - RMTeam, RMZ.cr
Description                    : Encoded By RMTeam, http://RMZ.cr
Writing application            : Lavf57.83.100
Writing library                : Lavf57.83.100
ErrorDetectionType             : Per level 1

Video
ID                             : 1
Format                         : HEVC
Format/Info                    : High Efficiency Video Coding
Format profile                 : Main@L3.1@Main
Codec ID                       : V_MPEGH/ISO/HEVC
Duration                       : 20 min 49 s
Width                          : 1 280 pixels
Height                         : 720 pixels
Display aspect ratio           : 16:9
Frame rate mode                : Variable
Original frame rate            : 23.976 FPS
Color space                    : YUV
Chroma subsampling             : 4:2:0
Bit depth                      : 8 bits
Writing library                : x265 2.5+27-0e168bdeb48b:[Linux][GCC 6.4.0][64 bit] 8bit+10bit+12bit
[...]

PR #216, opened 2018-01-22
Direct buffer video decoding

lukasf:
This PR further improves video decoding performance over #204 by eliminating all unneccessary copy operations. Decoder (or scaler, if required) will directly render into output buffer. BufferPool is used to keep memory usage stable (as opposed to allocating new memory for every frame).

If a frame size change is detected which exceeds the direct decoding buffer size, the solution automatically switches to scaler. I don't have any file to test this, but I think it could happen on broadcasts or maybe also incorrectly merged files.

One more thing I noticed: When we use Iyuv output, the DXVA hardware deinterlacer does not work, making interlaced mpeg2 files look a bit worse (e.g. the simpsons file). So I added code to enforce nv12 output for mpeg2. Hope this is a good idea.

### DIFF
--- a/FFmpegInterop/Source/FFmpegInteropMSS.cpp
+++ b/FFmpegInterop/Source/FFmpegInteropMSS.cpp
@@ -26,6 +26,7 @@
 #include "CritSec.h"
 #include "shcore.h"
 #include <mfapi.h>
+#include <dshow.h>
 
 extern "C"
 {
@@ -281,6 +282,34 @@ HRESULT FFmpegInteropMSS::CreateMediaStreamSource(IRandomAccessStream^ stream, b
 	return hr;
 }
 
+static int is_hwaccel_pix_fmt(enum AVPixelFormat pix_fmt)
+{
+	const AVPixFmtDescriptor *desc = av_pix_fmt_desc_get(pix_fmt);
+	return desc->flags & AV_PIX_FMT_FLAG_HWACCEL;
+}
+
+static AVPixelFormat get_format(struct AVCodecContext *s, const enum AVPixelFormat *fmt)
+{
+	AVPixelFormat result = (AVPixelFormat)-1;
+	AVPixelFormat format;
+	int index = 0;
+	do
+	{
+		format = fmt[index++];
+		if (format != -1 && result == -1 && !is_hwaccel_pix_fmt(format))
+		{
+			// take first non hw accelerated format
+			result = format;
+		}
+		else if (format == AV_PIX_FMT_NV12 && result != AV_PIX_FMT_YUVA420P)
+		{
+			// switch to NV12 if available, unless this is an alpha channel file
+			result = format;
+		}
+	} while (format != -1);
+	return result;
+}
+
 HRESULT FFmpegInteropMSS::InitFFmpegContext(bool forceAudioDecode, bool forceVideoDecode)
 {
 	HRESULT hr = S_OK;
@@ -398,6 +427,8 @@ HRESULT FFmpegInteropMSS::InitFFmpegContext(bool forceAudioDecode, bool forceVid
 
 				if (SUCCEEDED(hr))
 				{
+					avVideoCodecCtx->get_format = &get_format;
+
 					// initialize the stream parameters with demuxer information
 					if (avcodec_parameters_to_context(avVideoCodecCtx, avFormatCtx->streams[videoStreamIndex]->codecpar) < 0)
 					{
@@ -490,15 +521,11 @@ HRESULT FFmpegInteropMSS::InitFFmpegContext(bool forceAudioDecode, bool forceVid
 		}
 		if (mss)
 		{
+			mss->BufferTime = { 0 };
 			if (mediaDuration.Duration > 0)
 			{
 				mss->Duration = mediaDuration;
 				mss->CanSeek = true;
-			}
-			else
-			{
-				// Set buffer time to 0 for realtime streaming to reduce latency
-				mss->BufferTime = { 0 };
 			}
 
 			startingRequestedToken = mss->Starting += ref new TypedEventHandler<MediaStreamSource ^, MediaStreamSourceStartingEventArgs ^>(this, &FFmpegInteropMSS::OnStarting);
@@ -535,7 +562,7 @@ MediaThumbnailData ^ FFmpegInterop::FFmpegInteropMSS::ExtractThumbnail()
 			}
 
 
-			auto vector = ref new Array<uint8_t>(imageStream->attached_pic.data, imageStream->attached_pic.size);
+			auto vector = ArrayReference<uint8_t>(imageStream->attached_pic.data, imageStream->attached_pic.size);
 			DataWriter^ writer = ref new DataWriter();
 			writer->WriteBytes(vector);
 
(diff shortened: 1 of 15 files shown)


mcosmin222:
1) good job on the video part :)

2) it is probably better to do bigger PRs related to a specific feature rather than using smaller ones.

3) perhaps we could extend the BufferFactory to audio decoding as well?

mcosmin222:
I will test this entire PR in the weekend.

lukasf:
Regarding audio decoding: Yes I have been thinking about that as well. But the current implementation of the decoding loop does not allow us to use direct buffer for audio data. 

The current decoding loop always works with at least one packet as input, and expects one sample as output. But for audio data, one input packet can result in multiple output frames. So we concatenate all output frames resulting from a packet. This does not work with direct buffer approach. (And then we even decode more packets to fill up a minimum sample duration. Maybe that has been done to improve performance, not sure if it is really an improvement.)

Actually, I think that there is at least one bug in the decoding loop. We might have to change it anyways. Then we could implement it in a way that allows us to also try direct buffer for audio data, and see how that goes.

reego-fr:
I didn't notice the issue with deinterlacing but indeed there is something wrong with it. 
I think it is still working but not as well as on nv12 and I wonder if it could have something to do with chroma subsampling and chroma location. 
I was curious about trying these settings to see how it could influence rendering but never tried until now. It looks somewhat better but I am not 100% sure. I read somewhere that chroma location default for mpeg-2 frame and fields were different so I might not have tried with the good settings.

lukasf:
I am prett confident that what I see with nv12 is dxva hardware deinterlacing, while with Iyuv it is only (pretty poor) software deinterlacing. On fast scene changes there are the interlace stripes all over the image. That just does not happen with hw deinterlacing. And also during normal motion, the image is not clean, if you look closely.

I don't know much yet about chroma location, it is something I intended to look at some time. I hope that the defaults work well enough, but it might be possible to improve color accuracy/sharpness for some formats by fine tuning the chroma location. But I don't think that it causes the artifacts here. Both nv12 and iyuv have the same chroma plane bytes, they are just stored a little different (nv12 mixes the chroma planes while iyuv stores them separately).

reego-fr:
I agree iyuv and nv12 should be the same regarding chroma location, unless ffmpeg scaler does some sort of normalization which is very unlikely. 
Apart from scene changes you mentioned, I see mostly chroma artifacts this is why I investigated a bit on this part.

ramtinak:
Hi, I tested every branches that @lukasf was pushed.
It seems this branch works in phone perfectly for HEVC(720p) files!
There's no buffering that I mentioned in #208 , I used this by sets AudioDecode to false and VideoDecode to true, and works perfectly! but Just there's some issue, Audio is fine, but video is slow,
why is that?
I tested this by set both AudioDecode and VideoDecode to true, but this is not good at all!
What is the problem?

lukasf:
@ramtinak HEVC is a highly CPU intensive codec. Low and midrange phones are usually not capable of software decoding HEVC files. But newer phones like Lumia 550, 650 950 come with a hardware HEVC decoder. You can try my branch lukasf/hevc-passthrough-clean, to see if your phone is capable of hardware decoding HEVC files. If it is, then you should be able to get smooth playback (use VideoDecode=false).

ramtinak:
@lukasf I'v tried lukasf/hevc-passthrough-clean branch, when VideoDecode is false, just sound is playing and video is Black.

I think there most be a way to support HEVC in lower devices, because there is an media player(only) can play HEVC 720p in my phone smoothly.

lukasf:
@ramtinak What happens when you open that file directly in MediaPlayerElement, without FFmpegInterop? Also, does the file render on PC using FFmpegInterop with hevc passthrough or is it black as well?

reego-fr:
I saw that "slow video / ok audio" playback on some heavy files on my machines. Audio will eventually freeze at some time but it is surprising to see both stream will shift one from the other as if video presentation timestamps didn't count anymore while one would actually expect to get breaks in the audio when video frames are too late, or best, video frame skipping. 
I don't know if this behavior is due to multithreading or to MF pipeline but freeing resources certainly helped getting audio from these too heavy files. More testing might be needed but I consider it to be an edge case, I mean I don't expect 4k 60fps to be software decoded by my Atom CPU.

mcosmin222:
But is the same heavy file working on an average CPU? Like an i3? This may be a hardware issue or maybe there is a bottleneck somewhere.

lukasf:
I think this is just the way MF works when video decoding lags behind. I don't know why but MF just keeps playing audio even when video lags behind massively. Only at a very extreme point, playback finally stops and video catches up. I see the same kind of slow video on my older machine with 4K HEVC files, when playing them with the integrated Video or Pictures app. It is a 6 years old laptop but still quite powerful with Quad Core i7 @2GHz. But that is just not powerful enough for software decoding 4K HEVC files. There is no way how this can be fixed, if the CPU is just too slow.

The question for @ramtinak is more like why does hardware acceleration not work on his device, when a different player seems to use it successfully (at least, I don't think that other player uses sw decoding).

ramtinak:
@lukasf the answer to both of your question is playing audio without video(black page)

I really don't know how they use ffmpeg that can play hevc 720p in my phone, but I know minimum os version of this app is 10240 so they cannot use mediaplayerelement.

mcosmin222:
Are they "VLC"? if so, they are using a custom compoaition surface for their media player. Gives them better flexibility than using a builtin control.

The only othervway is to use a MediaElement, which you can also do with ffmpeg interop. Just set the mss as source for the media element object.

ramtinak:
@mcosmin222 No, VLC has very lags too! It's Cnx Player.
I know MediaElement! I always work with this one also test MediaPlayerElement.

lukasf:
@ramtinak If you can upload that file somewhere, I could check if there is a problem with that specific file. But it could also be that your phone's CPU does support HEVC HW decoding, but MS somehow did not enable that support (missing drivers or whatever). In that case, if Cnx Player uses custom implementation of HW decoding, they could enable it for their player. You will get black display when you try to hardware decode a file but the device does not support that file's specific format (e.g. if you passthrough Hi10P H264 files on PC, you will get black screen, because DXVA does not support Hi10P profile for H264).

For FFmpegInterop, it is not even possible to enable custom HW decoding, because of the limitation that MediaStreamSample can only be created from an IBuffer, but not from a D3D surface (which would be the output of HW acceleration).

You can vote for my UserVoice item to allow HW acceleration for MMS:

https://wpdev.uservoice.com/forums/110705-universal-windows-platform/suggestions/32994904-mediastreamsample-add-way-to-create-sample-from-d

ramtinak:
@lukasf It's not a specific file, I test every 720p x265 video in my phone, the only player that can play smoothly , it's what I mentioned.

Here is 3 of them:
http://dl5.film2movie.co/serial/Man%20With%20A%20Plan/S02/Man.With.A.Plan.S02E03.720p.x265.HEVC.Film2Movie_CO.mkv

http://dl6.film2movie.co/user/erfan/serial/The%20Resident/S01/The.Resident.S01E03.720p.x265.HEVC.Film2Movie_CO.mkv

http://dl5.film2movie.co/serial/Black%20Lightning/S01/Black.Lightning.S01E01.720p.x265.HEVC.Film2Movie_CO.mkv

I don't think my phone has HEVC HW decoding, because if it does, other players that supports HW decoding would play these files, not disable it for HEVC files.

And something else that I forgot to mentioned, I have a 720p mkv file (codec id = V_VP9) It's kind of weird, because like 720p HEVC files, only Cnx Player could plays it smoothly!

here: 
http://dl.subrica.download/musicvideos/Era%20Istrefi/Era%20Istrefi%20-%20Bonbon%20(720)(subrica).mkv

PR #220, opened 2018-02-13
Filters

mcosmin222:
I created a framework for inserting filter graphs into Uncompressed sample providers and implemented an example audio filter class (which works out of the box).

You can pass filters using the configurations described in the link bellow by creating AvEffectDefinition objects

For example, you can insert an aecho effect like so

new AvEffectDefinition("aecho", "aecho=0.8:0.88:60:0.4")

Effects can be enabled or disabled during playback at will.
I have tested this branch a little, it has no obvious memory leaks or hard hitting errors.

I haven't done anything for videos yet (should be roughly similar to the existing AudioEffect class, but with different configurations for buffer source and resampler filter), and I am not sure if it is even possible for compressed samples.

https://ffmpeg.org/ffmpeg-filters.html#Audio-Filters

PS: I am not sure why the entire history of the repository is visible here, only the last 2 commits matter.

PS2: I know filters have several ways of initialization, but i find the string based approach the best, and it seems to work every time, and is easiest and most versatile to implement. The context options API doesn't seem to work every time. I only used it for the abuffer source because that filter has many options that would be harder to express as strings.


lukasf:
@mcosmin222 I know this problem. I solve it this way: Create a new clean branch, based off the branch where you want to merge (in this case, master). Checkout that branch, then use git bash: git merge --squash <otherBranch> (<otherBranch> could then be "lukasf/filters" or just a different branch in the same repo).

mcosmin222:
This was my mistake. The PR should have been in your parallel repo, not here. I guess the coxe was eager to come here :)

lukasf:
Oh ok I see :)

mcosmin222:
We can move over there :)

PR #233, opened 2018-07-06
Revert "Skip a few broken packets. Disable stream processing once thr…

dcarpente:
…eshold has been exceeded."

This reverts commit 83bb3cc1ebed271d6b8e0fa165bc51afccb3f4f9.

This change made the decoder unable to play after hitting EOS for the first time


lukasf:
Generally it would be better to repair a broken feature than just remove it. I think you just need to set isEnabled=true after a seek operation.

You can also take a look at my fork of the lib. It is a much improved version with lots of bug fixes and new features.

dcarpente:
@lukasf, thanks for your comment. Based on your feedback, does this look better?

Of course I'd like to fix a broken feature instead of removing it, but sometimes that's the easiest fix with limited information. :)

lukasf:
Yes this looks fine!

PR #263, opened 2019-09-17
Add subtitle stream support

brbeec:
This change adds support for SRT, SSA, PGS, and VobSub subtitle streams. A SwitchStreamsRequested event handler was implemented to allow enabling/disabling the subtitle stream during playback.


FFmpegInterop/FFmpegInteropMSS.cpp
```diff
@@ -388,33 +404,17 @@ HRESULT FFmpegInteropMSS::InitFFmpegContext(bool forceAudioDecode, bool forceVid
...
-
 				// Create audio stream descriptor from parameters
-				hr = CreateAudioStreamDescriptorFromParameters(avAudioCodecParams);
+				hr = CreateAudioStreamDescriptorFromParameters(avFormatCtx->streams[audioStreamIndex]->codecpar);
 				if (SUCCEEDED(hr))
 				{
 					hr = audioSampleProvider->AllocateResources();
 					if (SUCCEEDED(hr))
 					{
+						audioSampleProvider->EnableStream();
```
isuru-c-p: >audioSampleProvider->EnableStream(); [](start = 6, length = 36)

Shouldn't "audioStreamSelected" be set to true for this case as well?

isuru-c-p:
> {

Need to initialize m_subtitleStreamIndex as well.

---
Refers to: FFmpegInterop/FFmpegReader.cpp:28 in dd5a2e6. [](commit_id = dd5a2e6dd3afdd998cada6aa01dc04bba44b4cf3, deletion_comment = False)

PR #268, opened 2020-02-12
Convert FFmpegInterop from C++/CX to C++/WinRT

brbeec:
This change is a rewrite of FFmpegInterop that converts it from C++/CX to C++/WinRT.

The following is a list of notable changes:
* HRESULTs -> Exceptions
  * We now use exceptions everywhere internally and convert to HRESULTs at the ABI boundary. WIL error handling macros are used for control flow and error propagation.
* Tracing
  * New TraceLogging activity classes were added for the MSS event handlers. These classes provide enter/exit scope tracing as well as error tracing via the WIL error macros and the wil::ThreadFailureCache. One caveat with the wil::ThreadFailureCache is that we only get error information if the exception was observed by a WIL error handling macro. In practice this means that we won't have tracing for exceptions thrown from STL calls.
  * Added new TraceLogging events to help aid debugging.
* N stream support
  * Added support for any number of audio/video/subtitle streams. We previously only supported one of each type.
* Added custom buffer class to minimize sample copies internally
  * The custom buffer class implements IMarshal to support marshaling the buffer cross process. This is needed for media source app service scenarios.
* RAII resource wrappers
  * Added new RAII resource wrapper classes for smartly managing FFmpeg resources.
* Expose more media types
  * Added support for exposing more media types that can be decoded by MF
* Switched to push model for subtitle streams
  * Subtitle stream samples may be sparse within a file. The current pull model where we scan until we find the next sample for the stream or hit EOF has perf issues: scanning could take a long time and also queue a bunch of samples for other streams in the process. To address these issues subtitle streams have been switched to a push model. When we receive a sample request for a subtitle stream, we defer the request until a sample for the stream becomes available while scanning for the audio/video streams. 
* Metadata
  * Removed MediaThumbnailData class. Use MSS Thumbnail property instead.
  * Added skeleton code for future metadata changes.
* Updated MediaPlayer test apps
  * Converted C++ MediaPlayer test app from C++/CX to C++/WinRT
  * Added support for file activation.


FFmpegInterop/FFmpegInteropLogging.h
```diff
@@ -17,21 +17,30 @@
...
 #pragma once
-#include "ILogProvider.h"
 
-namespace FFmpegInterop
+#include "FFmpegInteropLogging.g.h"
+
+namespace winrt::FFmpegInterop::implementation
 {
-	public ref class FFmpegInteropLogging sealed
+	// TODO: Update for trace logging
```
dcarpente: Is this still TODO? We should either track with internal work item or remove

FFmpegInterop/H264SampleProvider.cpp
```diff
@@ -19,54 +19,136 @@
...
+		m_profile = data[1];
+
+		uint8_t profileCompatibility{ data[2] };
+		m_constraintSet1 = (profileCompatibility & (1 << 6)) != 0;
 	}
 
-	return hr;
-}
\ No newline at end of file
+	AVCPictureParamterSetParser::AVCPictureParamterSetParser(_In_reads_(dataSize) const uint8_t* data, _In_ int dataSize)
```
dcarpente: "Paramter" Typo

FFmpegInterop/H264SampleProvider.cpp
```diff
@@ -19,54 +19,136 @@
...
+	{
+		// TODO: Implement
```
dcarpente: It seems like there's some missing work for this? (Also, TODO on MF_MT_VIDEO_H264_NO_FMOASO.) Put some information about what's planned here under "Expose more media types" in your description.

FFmpegInterop/FFmpegInteropBuffer.h
```diff
@@ -0,0 +1,78 @@
...
+
+		// IBuffer
+		uint32_t Capacity() const noexcept;
+		uint32_t Length() const noexcept;
+		void Length(_In_ uint32_t length);
+
+		// IBufferByteAccess
+		STDMETHODIMP Buffer(_Outptr_ uint8_t** buf) noexcept;
+
+		// IMarshal
```
dcarpente: Why does this implement IMarshal, is it used cross process? Can there be some info about this in the description?

FFmpegInterop/Metadata.cpp
```diff
@@ -0,0 +1,56 @@
...
+	void PopulateMSSMetadata(_In_ const MediaStreamSource& mss, _In_ const AVDictionary* metadata)
+	{
+		// Get the MSS property handler
+		com_ptr<IMFGetService> mediaSource;
+		THROW_IF_FAILED(mss.as<IMFGetService>()->GetService(MF_MEDIASOURCE_SERVICE, __uuidof(mediaSource), mediaSource.put_void()));
+
+		com_ptr<IPropertyStore> propHandler;
+		THROW_IF_FAILED(mediaSource->GetService(MF_PROPERTY_HANDLER_SERVICE, __uuidof(propHandler), propHandler.put_void()));
+
+		// TODO: Populate the property handler with the metadata
```
dcarpente: Remove TODO comment or add some explanation about future work in description

FFmpegInterop/StreamFactory.cpp
```diff
@@ -0,0 +1,366 @@
+//*****************************************************************************
+//
+//	Copyright 2015 Microsoft Corporation
```
dcarpente: Some files have Copyright 2015 instead of 2019, is there a difference or is it just a mistake?

FFmpegInterop/Tracing.h
```diff
@@ -0,0 +1,106 @@
...
+//
+//*****************************************************************************
+
+#pragma once
+
+namespace winrt::FFmpegInterop::implementation
+{
+	TRACELOGGING_DECLARE_PROVIDER(g_FFmpegInteropProvider);
+
+	// TODO: Replace these macros with WIL TraceLogging macros if they're ever made public
```
dcarpente: I know this work probably isn't planned, just remove the TODO. E.g. "These macros should be replaced with WIL TraceLogging macros if they're ever made public."

FFmpegInterop/FFmpegInteropLogging.h
brbeec: Yes, I'd like to get TraceLogging working with this so we can get additional tracing from FFmpeg in ETLs. I'll file a task for it.

Tests/TestCreateFFmpegInteropMSSFromStream.cs
```diff
@@ -30,216 +31,170 @@ namespace UnitTest.Windows
...
-
-            mss = null;
-            Assert.IsNull(mss);
-            Assert.IsNotNull(Originalmss);
-
-            OriginalFFmpegMSS = null;
-            Originalmss = null;
-            Assert.IsNull(OriginalFFmpegMSS);
-            Assert.IsNull(Originalmss);
+            // TODO: Verify expected stream types (aac, h264)
```
dcarpente: Add explanation about future change or remove

Tests/TestCreateFFmpegInteropMSSFromUri.cs
```diff
@@ -18,177 +18,163 @@
...
+            MediaStreamSource mss = CreateMSSFromUri(Constants.StreamingUriSource, null);
 
             // Based on the provided media, check if the following properties are set correctly
-            Assert.AreEqual(true, mss.CanSeek);
-            Assert.AreNotEqual(0, mss.BufferTime.TotalMilliseconds);
+            Assert.IsTrue(mss.CanSeek);
+            Assert.IsTrue(mss.BufferTime.TotalMilliseconds > 0);
             Assert.AreEqual(Constants.StreamingUriLength, mss.Duration.TotalMilliseconds);
+
+            // TODO: Verify expected stream types (aac, h264)
```
dcarpente: Multiple TODO for verifying expected stream types in these tests.

FFmpegInterop/H264SampleProvider.cpp
brbeec: H264 playback is working. The remaining work is just for MF_MT_VIDEO_H264_NO_FMOASO which is an H264 decoder optimization. From mfapi.h: "MF_MT_VIDEO_H264_NO_FMOASO set to non-zero (true) means external users/apps know that H.264 input video bitstream has no FMO/ASO enabled, that is, even if the bitstream has baseline profile and constraint_set1_flag equal to 0, the bitstream shall not have FMO/ASO then H.264 decoder uses DXVA decoding and doesn't fall back to software decoding it improves power consumption, memory usage, performance and user experiences (without unnecessary glitches on low end devices)."

FFmpegInterop/StreamFactory.cpp
brbeec: I updated the date when copying to new files.

FFmpegInterop/FFmpegInteropBuffer.h
brbeec: Yes, in app service scenarios the buffer is marshaled cross proc. I'll add a comment.

brbeec: Yes, in media source app service scenarios the buffer is marshaled cross proc. I'll add a comment about it.

FFmpegInterop/H264SampleProvider.cpp
brbeec: Fixed.

FFmpegInterop/Tracing.h
brbeec: Removed.

FFmpegInterop/H264SampleProvider.cpp
brbeec: I added a bitstream reader for SPS/PPS parsing and implemented the MF_MT_VIDEO_H264_NO_FMOASO logic.

FFmpegInterop/BitstreamReader.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+	void BitstreamReader::SkipN(_In_ uint32_t numBits)
+	{
+		// Make sure we have enough data left to fulfill the request
+		THROW_HR_IF(MF_E_INVALID_POSITION, numBits > BitsRemaining());
+
+		auto [numBytesToSkip, numBitsToSkip] = div(numBits, BITS_PER_BYTE);
+
+		m_byteIndex += numBytesToSkip;
+		m_bitIndex += numBitsToSkip;
```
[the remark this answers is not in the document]
brbeec: Yeah this logic is wrong. The conditional should be:

if (m_bitIndex >= BITS_PER_BYTE)
{
    m_byteIndex++;
    m_bitIndex -= BITS_PER_BYTE;
}

FFmpegInterop/FFmpegInteropBuffer.cpp
```diff
@@ -0,0 +1,184 @@
...
+	STDMETHODIMP FFmpegInteropBuffer::DisconnectObject(_Reserved_ DWORD dwReserved) noexcept
+	{
+		RETURN_IF_FAILED(InitMarshaler());
+		RETURN_IF_FAILED(m_marshaler->DisconnectObject(dwReserved));
+		return S_OK;
+	}
+
+	HRESULT FFmpegInteropBuffer::InitMarshaler() noexcept
+	{
+		if (m_marshaler == nullptr)
```
isuru-c-p, line 172: >m_marshaler == nullptr [](start = 6, length = 22)

You will need to use InterlockedCompareExchangePointer here to ensure that this read is synchronized with the write within the if statement (i.e InterlockedCompareExchangePointer(m_marshaler.addressof(), nullptr, nullptr) == nullptr). Otherwise, on some platforms, this check could evaluate to false while another thread is still writing to m_marshaler, causing a subsequent method call on an invalid m_marshaler pointer.

FFmpegInterop/FFmpegInteropMSS.cpp
```diff
@@ -18,1020 +18,432 @@
...
-		// Find the video stream and its decoder
-		AVCodec* avVideoCodec = nullptr;
-		videoStreamIndex = av_find_best_stream(avFormatCtx, AVMEDIA_TYPE_VIDEO, -1, -1, &avVideoCodec, 0);
-		if (videoStreamIndex != AVERROR_STREAM_NOT_FOUND && avVideoCodec)
+		THROW_HR_IF_FFMPEG_FAILED(avformat_find_stream_info(m_formatContext.get(), nullptr));
+
+		bool hasVideo{ false };
+		bool hasAudio{ false };
+
+		for (unsigned int i{ 0 }; i < m_formatContext->nb_streams; i++)
```
isuru-c-p, line 203: >for (unsigned int i{ 0 }; i < m_formatContext->nb_streams; i++) [](start = 2, length = 63)

nit: There's a potential regression here in terms of default stream selection. Previously, we were calling av_find_best_stream to find the indices of the default audio / video streams to select. av_find_best_stream has logic to select the best default stream based on bitrate / disposition, whereas we're now selecting the first audio + video stream that we enumerate. We could potentially retain the default stream selection behavior for audio / video streams by adding logic after this loop to perform selection based on the results from av_find_best_stream.

FFmpegInterop/FFmpegInteropMSS.cpp
```diff
@@ -18,1020 +18,432 @@
...
-	in.QuadPart = pos;
-	ULARGE_INTEGER out = { 0 };
+		lock_guard<mutex> lock{ m_lock };
 
-	if (FAILED(pStream->Seek(in, whence, &out)))
-	{
-		return -1;
-	}
+		// Unregister event handlers
+		// This is critcally important to do for the media source app service scenario! If we don't unregister these event handlers, then
```
isuru-c-p, line 431: >critcally [](start = 13, length = 9)

Typo here and below.

FFmpegInterop/FFmpegInteropMSS.cpp
```diff
@@ -18,1020 +18,432 @@
...
+		// Setup FFmpeg custom IO to access file as stream. This is necessary when accessing any file outside of app installation directory and appdata folder.
+		// Credit to Philipp Sch http://www.codeproject.com/Tips/489450/Creating-Custom-FFmpeg-IO-Context
+		constexpr int c_ioBufferSize = 16 * 1024;
+		AVBlob_ptr ioBuffer{ av_malloc(c_ioBufferSize) };
+		THROW_IF_NULL_ALLOC(ioBuffer);
 
-HRESULT FFmpegInteropMSS::InitFFmpegContext(bool forceAudioDecode, bool forceVideoDecode)
-{
-	HRESULT hr = S_OK;
+		m_ioContext.reset(avio_alloc_context(reinterpret_cast<unsigned char*>(ioBuffer.get()), c_ioBufferSize, 0, m_fileStream.get(), FileStreamRead, 0, FileStreamSeek));
```
isuru-c-p, line 148: >0 [](start = 144, length = 1)

nit: Use nullptr for this arg as its a function pointer.

FFmpegInterop/SampleProvider.cpp
```diff
@@ -0,0 +1,415 @@
...
+			videoEncProp.Bitrate(static_cast<uint32_t>(codecPar->bit_rate));
+
+			if (codecPar->sample_aspect_ratio.num > 0 && codecPar->sample_aspect_ratio.den != 0)
+			{
+				MediaRatio pixelAspectRatio{ videoEncProp.PixelAspectRatio() };
+				pixelAspectRatio.Numerator(codecPar->sample_aspect_ratio.num);
+				pixelAspectRatio.Denominator(codecPar->sample_aspect_ratio.den);
+			}
+
+			if (m_stream->avg_frame_rate.num != 0 || m_stream->avg_frame_rate.den != 0)
```
isuru-c-p, line 93: >|| [](start = 41, length = 2)

Should this be &&?

FFmpegInterop/SampleProvider.cpp
```diff
@@ -0,0 +1,415 @@
...
+				// Set the format user data
+				IVideoEncodingProperties2 videoEncProp2{ encProp.as<IVideoEncodingProperties2>() };
+				videoEncProp2.SetFormatUserData({ codecPar->extradata, codecPar->extradata + codecPar->extradata_size });
+			}
+
+			MediaPropertySet properties{ videoEncProp.Properties() };
+
+			properties.Insert(MF_MT_ALL_SAMPLES_INDEPENDENT, PropertyValue::CreateUInt32(false));
+			properties.Insert(MF_MT_COMPRESSED, PropertyValue::CreateUInt32(true));
+			properties.Insert(MF_MT_PIXEL_ASPECT_RATIO, PropertyValue::CreateUInt64(Pack2UINT32AsUINT64(1, 1)));
```
isuru-c-p, line 111: >properties.Insert(MF_MT_PIXEL_ASPECT_RATIO, PropertyValue::CreateUInt64(Pack2UINT32AsUINT64(1, 1))); [](start = 3, length = 100)

It looks like this would override the pixelAspectRatio set above - can this be removed?

FFmpegInterop/FFmpegInteropMSS.h
```diff
@@ -17,148 +17,56 @@
...
+		void OnSampleRequested(_In_ const Windows::Media::Core::MediaStreamSource& sender, _In_ const Windows::Media::Core::MediaStreamSourceSampleRequestedEventArgs& args);
+		void OnSwitchStreamsRequested(_In_ const Windows::Media::Core::MediaStreamSource& sender, _In_ const Windows::Media::Core::MediaStreamSourceSwitchStreamsRequestedEventArgs& args);
+		void OnClosed(_In_ const Windows::Media::Core::MediaStreamSource& sender, _In_ const Windows::Media::Core::MediaStreamSourceClosedEventArgs& args);
 
-		MediaStreamSource^ mss;
-		EventRegistrationToken startingRequestedToken;
-		EventRegistrationToken sampleRequestedToken;
-		EventRegistrationToken switchStreamsRequestedToken;
+		std::mutex m_lock;
+		Windows::Media::Core::MediaStreamSource m_mss; // We hold a circular reference to the provided MSS which we break when the Closed event is fired
```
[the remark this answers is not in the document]
isuru-c-p: On closer inspection, it looks like this is specific to the new implementation as we now explicitly hold a strong reference to FFmpegInteropMSS in the delegates:

`// Register event handlers. The delegates hold strong references to tie the lifetime of this object to the MSS.`
`m_startingEventToken = m_mss.Starting({ get_strong(), &FFmpegInteropMSS::OnStarting });`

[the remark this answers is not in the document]
brbeec: Lukas is correct that the events were weak references before so there was no circular reference. I made them strong references using get_strong() so that the FFmpegInteropMSS is guaranteed to be kept alive as long as it has events registered with the MSS.

There's a good C++/WinRT article covering this topic:
https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/weak-references

FFmpegInterop/FFmpegInteropBuffer.cpp
brbeec: Nice catch! I modeled this code after other [Microsoft sample projects](https://github.com/search?q=org%3Amicrosoft+RoGetBufferMarshaler&type=Code) so we should fix this issue in those projects too.

I'm actually going to switch to using std::call_once() for lazy initialization here. Calling com_ptr::put_void() is technically valid here because we'll only update m_marshaler if it's null, however, this usage still triggers the assert in com_ptr::put() which is problematic on debug builds.

FFmpegInterop/FLACSampleProvider.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+	}
+
+	tuple<IBuffer, int64_t, int64_t, map<GUID, IInspectable>> FLACSampleProvider::GetSampleData()
+	{
+		if (m_isDiscontinuous)
+		{
+			// On discontinuity send the codec private data containing the FLAC STREAMINFO block as the first sample to the decoder.
+			// We need to prepend the FLAC STREAMINFO block header as FFmpeg strips this out.
+			constexpr uint32_t FLAC_STREAMINFO_SIZE{ 34 };
```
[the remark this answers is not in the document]
brbeec: Hmm this is the same behavior as the in-box MKV source and it seemed to be working.

FFmpegInterop/SampleProvider.cpp
```diff
@@ -0,0 +1,415 @@
...
+			TraceLoggingValue(m_stream->index, "StreamId"));
+
+		WINRT_ASSERT(m_isSelected);
+		m_isSelected = false;
+		Flush();
+	}
+
+	void SampleProvider::OnSeek(_In_ int64_t seekTime) noexcept
+	{
+		m_nextSamplePts = seekTime;
```
isuru-c-p, line 304: >m_nextSamplePts = seekTime; [](start = 2, length = 27)

The seekTime provided here is in AV_TIME_BASE units, whereas m_nextSamplePts is assumed to be in stream specific time_base units in SampleProvider::GetSample. I'm not sure how common it is for the stream specific time_base units to not be equivalent to AV_TIME_BASE units, but we should probably be more robust here and handle that case.

FFmpegInterop/SampleProvider.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+		return { make<FFmpegInteropBuffer>(move(packet)), pts, dur, move(properties) };
+	}
+
+	AVPacket_ptr SampleProvider::GetPacket()
+	{
+		// Continue reading until there is an appropriate packet in the stream
+		while (m_packetQueue.empty())
+		{
+			m_reader.ReadPacket();
```
isuru-c-p, line 407: >m_reader [](start = 3, length = 8)

nit: I know you just followed the pre-existing design here, but IMO this approach of plumbing the reader into each sample provider and having the sample providers call ReadPacket() isn't very clean. Specifically because calling ReadPacket() on the reader from a sample provider would also impact the state (queued packets) of other sample providers. It would be cleaner if we had this logic within the parent FFmpegInteropMSS class instead (i.e in FFmpegInteropMSS::OnSampleRequested, we could check whether the sample provider has queued packets, and if not, perform packet reads until we've queued a packet for it). We can probably leave this as it is for now, but it's something to consider as a future cleanup task.

isuru-c-p: Another issue which would be alleviated by having this logic in the parent FFmpegInteropMSS class is the fact that there is an implicit contract here that m_reader.ReadPacket will eventually either fail (throw) or cause a packet to be added to this sample provider's packet queue. This is kind of fragile - there is already a case where the sample provider won't add a packet to its queue when the reader calls QueuePacket() on it (when the stream is not selected). We currently have a mutex in the parent FFmpegInteropMSS class to prevent DeselectStream from being called on a sample provider while this loop is executing - it would be easier to reason about this and prevent future regressions in this area if this loop was also in FFmpegInteropMSS.

FFmpegInterop/UncompressedVideoSampleProvider.cpp
```diff
@@ -18,140 +18,111 @@
...
-			m_pAvCodecCtx->width,
-			m_pAvCodecCtx->height,
-			AV_PIX_FMT_NV12,
-			SWS_BICUBIC,
-			NULL,
-			NULL,
-			NULL);
-
-		if (m_pSwsCtx == nullptr)
+		if (m_codecContext->pix_fmt != AV_PIX_FMT_NV12)
```
isuru-c-p, line 33: >m_codecContext->pix_fmt != AV_PIX_FMT_NV12 [](start = 6, length = 42)

The MF pipeline supports more uncompressed video formats than just NV12 - perhaps we could file a task to track expanding media type support in this sample provider in future.

FFmpegInterop/UncompressedSampleProvider.cpp
```diff
@@ -19,105 +19,65 @@
...
 			{
-				framePts = m_pAvFrame->pts;
-				frameDuration = m_pAvFrame->pkt_duration;
+				// The decoder needs more data to produce a frame
+				TraceLoggingWrite(g_FFmpegInteropProvider, "DecoderNeedsMoreInput", TraceLoggingLevel(TRACE_LEVEL_VERBOSE), TraceLoggingPointer(this, "this"),
+					TraceLoggingValue(m_stream->index, "StreamId"));
+				continue;
 			}
-			fGotFrame = true;
+			THROW_HR_IF_FFMPEG_FAILED(decodeResult);
```
isuru-c-p, line 76: >THROW_HR_IF_FFMPEG_FAILED(decodeResult); [](start = 3, length = 40)

FFmpegInterop previously had logic to allow a certain number of decode failures before failing out. If we don't think that it's worth maintaining that behavior, we should probably document this change for consumers (just in the commit message would be ok for now). I also noticed another behavior change in that we now fail playback when failing to get a sample for a stream, whereas previously the stream was disabled and playback continued.

FFmpegInterop/Reader.cpp
```diff
@@ -0,0 +1,55 @@
...
+	Reader::Reader(_In_ AVFormatContext* formatContext, _In_ const map<int, SampleProvider*>& streamMap) :
+		m_formatContext(formatContext),
+		m_streamIdMap(streamMap)
+	{
+
+	}
+
+	void Reader::ReadPacket()
+	{
+		AVPacket_ptr packet{ av_packet_alloc() };
```
isuru-c-p, line 36: >AVPacket_ptr packet{ av_packet_alloc() }; [](start = 2, length = 41)

Performing a heap allocation each time a packet is read is not ideal for perf, but it looks like fixing this would require a number of changes since the QueuePacket() method on the sample provider takes in a unique_ptr, which prevents us from using stack space here (like the old implementation of FFmpegReader::ReadPacket) and it also prevents us from making the AVPacket poolable. We could potentially track this as a future perf improvement (e.g by creating a pooled version of AVPacket_ptr which QueuePacket accepts as an argument). There might also be no point in optimizing this if the underlying ffmpeg demuxer doesn't pool buffer allocations.

FFmpegInterop/NALUSampleProvider.cpp
```diff
@@ -0,0 +1,252 @@
...
+		SampleProvider::SetEncodingProperties(encProp, setFormatUserData);
+
+		VideoEncodingProperties videoEncProp{ encProp.as<VideoEncodingProperties>() };
+		videoEncProp.ProfileId(m_stream->codecpar->profile);
+
+		MediaPropertySet videoProp{ videoEncProp.Properties() };
+		videoProp.Insert(MF_MT_MPEG2_LEVEL, PropertyValue::CreateUInt32(static_cast<uint32_t>(m_stream->codecpar->level)));
+		videoProp.Insert(MF_NALU_LENGTH_SET, PropertyValue::CreateUInt32(static_cast<uint32_t>(true)));
+
+		videoProp.Insert(MF_MT_MPEG_SEQUENCE_HEADER, PropertyValue::CreateUInt8Array({ m_spsPpsData.data(), m_spsPpsData.data() + m_spsPpsData.size() }));
```
isuru-c-p, line 98: >MF_MT_MPEG_SEQUENCE_HEADER [](start = 19, length = 26)

This should only be set if m_spsPpsData is not empty

FFmpegInterop/FFmpegInteropMSS.h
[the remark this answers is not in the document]
lukasf: Actually, after I wrote my comment, I thought that it would be cool if this was a strong reference, for the reasons @brbeec mentioned. I will try to get this change also into our fork, just need to find a way to do it in C++/CX.

FFmpegInterop/SampleProvider.cpp
brbeec: That's what was there before, but yeah I think you're right.

FFmpegInterop/SampleProvider.cpp
brbeec: Yeah we can revisit this in the future. One complication is that the FFmpegInteropMSS doesn't know ahead of time how many packets a sample provider will need to produce a sample.  The current design nicely encapsulates special case logic in the sample provider.

Examples:
- Uncompressed sample providers need a varying number of packets to produce a sample. At EOF when the decoder enters draining mode the sample provider may still be able to generate samples without any additional packet input.
- The subtitle sample provider defers sample requests when its packet queue is empty.

FFmpegInterop/UncompressedVideoSampleProvider.cpp
brbeec: I filed a task.

FFmpegInterop/Reader.cpp
brbeec: We can revisit pooling AVPackets in the future. For now I added an optimization to set AVDISCARD_ALL on deselected streams so that we don't waste cycles allocating and reading packets that we're just going to immediately throw away.

Tests/TestCreateFFmpegInteropMSSFromStream.cs
brbeec: I'll handle test updates in a future change.

Tests/TestCreateFFmpegInteropMSSFromUri.cs
brbeec: I'll handle test updates in a future change.

FFmpegInterop/SampleProvider.cpp
```diff
@@ -0,0 +1,414 @@
...
+
+namespace winrt::FFmpegInterop::implementation
+{
+	SampleProvider::SampleProvider(_In_ AVStream* stream, _In_ Reader& reader) :
+		m_stream(stream),
+		m_reader(reader)
+	{
+		if (m_stream->start_time != AV_NOPTS_VALUE)
+		{
+			m_startOffset = m_stream->start_time;
```
isuru-c-p, line 37: Wouldn't you need to set m_startOffset to the start_time in the AVFormatContext instead? Otherwise, you'd rebase all streams to start at 0 and re-introduce the bug which was fixed in b4a95dc.

FFmpegInterop/NALUSampleProvider.cpp
```diff
@@ -0,0 +1,255 @@
...
+			}
+			else
+			{
+				naluLength = GetAVCNaluLength(packet->data + i, packet->size - i, m_naluLengthSize);
+			}
+
+			if (copySpsPpsData)
+			{
+				// Check if this NALU is an AUD (access unit delimiter). The AUD must be the first NALU in the sample.
+				if (naluLength < naluPrefixLength || packet->data[naluPrefixLength] != NALU_TYPE_AUD)
```
isuru-c-p, line 176: >naluLength < naluPrefixLength [](start = 8, length = 29)

Shouldn't this be "naluLength == 0"? The length of an AUD NALU can be less than the prefix length - the payload just contains the NALU header (1 byte) + primary_pic_type (3 bits) + padding for byte alignment (5 bits).

FFmpegInterop/NALUSampleProvider.cpp
```diff
@@ -0,0 +1,255 @@
...
+					copySpsPpsData = false;
+				}
+			}
+
+			if (writeToBuf)
+			{
+				// Write the NALU start code and data
+				const uint8_t* naluData{ packet->data + i + naluPrefixLength };
+				buf.insert(buf.end(), begin(NALU_START_CODE), end(NALU_START_CODE));
+				buf.insert(buf.end(), naluData, naluData + naluLength);
```
isuru-c-p, line 193: >buf.insert(buf.end(), naluData, naluData + naluLength); [](start = 4, length = 55)

Will this throw if naluLength is 0? If so, we should probably have an explicit check here to skip the insert call since we wouldn't want the presence of empty NALUs to cause fatal failures.

FFmpegInterop/NALUSampleProvider.cpp
```diff
@@ -0,0 +1,255 @@
...
+			}
+			else if (!m_isBitstreamAnnexB)
+			{
+				// Replace the NALU length with the NALU start code
```
isuru-c-p, line 197: >// [](start = 4, length = 2)

nit: Might be good to assert that (m_naluLengthSize == sizeof(NALU_START_CODE)) here (it improves readability and could prevent future regressions).

FFmpegInterop/NALUSampleProvider.cpp
```diff
@@ -0,0 +1,255 @@
...
+		vector<uint32_t> spsPpsNaluLengths;
+
+		for (uint32_t i{ 0 }; i < m_dataSize;)
+		{
+			uint32_t naluLength{ GetAnnexBNaluLength(m_data + i, m_dataSize - i) };
+
+			// Save the NALU length
+			spsPpsNaluLengths.push_back(naluLength);
+
+			i += naluLength;
```
isuru-c-p, line 250: >i += naluLength; [](start = 3, length = 16)

Shouldn't this be: i+= (sizeof(NALU_START_CODE) + naluLength)?

isuru-c-p: Actually, the naluLength initializer should be updated to add sizeof(NALU_START_CODE) since NALUSampleProvider::TransformSample assumes that the NALU lengths in m_spsPpsNaluLengths include the size of the start code.

---
In reply to: [406526441](https://github.com/microsoft/FFmpegInterop/pull/268#discussion_r406526441) [](ancestors = 406526441)

FFmpegInterop/NALUSampleProvider.cpp
```diff
@@ -0,0 +1,255 @@
...
+		m_data(data),
+		m_dataSize(dataSize)
+	{
+		// Validate parameters
+		WINRT_ASSERT(m_data != nullptr);
+		THROW_HR_IF(MF_E_INVALID_FILE_FORMAT, m_dataSize < sizeof(NALU_START_CODE));
+		THROW_HR_IF(MF_E_INVALID_FILE_FORMAT, !equal(begin(NALU_START_CODE), end(NALU_START_CODE), m_data));
+	}
+
+	tuple<vector<uint8_t>, vector<uint32_t>> AnnexBParser::GetSpsPpsData() const
```
isuru-c-p, line 238: >GetSpsPpsData [](start = 56, length = 13)

nit: This method implementation isn't SPS / PPS data specific, so perhaps it would be better to use a more generic name here (e.g GetNaluData()). This applies to the spsPpsData / spsPpsNaluLengths local variables in this method as well.

FFmpegInterop/H264SampleProvider.cpp
```diff
@@ -18,55 +18,209 @@
...
+	{
+		// Reserve estimated space now to minimize reallocations
+		spsPpsNaluLengths.reserve(spsPpsNaluLengths.size() + parameterSetCount);
+
+		// Parse the parameter sets and convert the SPS/PPS data to Annex B format
+		uint32_t pos{ 0 };
+		for (uint8_t i{ 0 }; i < parameterSetCount; i++)
+		{
+			// Get the NALU length
+			uint32_t naluLength{ GetAVCNaluLength(data + pos, dataSize - pos, sizeof(uint16_t)) };
```
isuru-c-p, line 129: >dataSize - pos [](start = 53, length = 14)

Need to ensure that pos doesn't exceed dataSize prior to this call (to prevent an underflow here and a subsequent OOB read in GetAVCNaluLength).

FFmpegInterop/H264SampleProvider.cpp
```diff
@@ -18,55 +18,209 @@
...
+			pos += sizeof(uint16_t);
 
-	if (!m_pAvCodecCtx || (m_pAvCodecCtx->extradata == nullptr && m_pAvCodecCtx->extradata_size < 8))
+			// Write the NALU start code and SPS/PPS data to the buffer
+			spsPpsData.insert(spsPpsData.end(), begin(NALU_START_CODE), end(NALU_START_CODE));
+			spsPpsData.insert(spsPpsData.end(), data + pos, data + pos + naluLength);
```
isuru-c-p, line 134: >data + pos + naluLength [](start = 51, length = 23)

Need a bounds check to ensure that (data + pos + naluLength) does not extend beyond the end of the buffer.

FFmpegInterop/H264SampleProvider.cpp
```diff
@@ -18,55 +18,209 @@
...
+		uint32_t pos{ 5 };
+
+		// Scan the SPS NALUs and check if any have a non-constrained baseline
+		bool foundNonConstrainedBaselineSPS{ false };
+		uint8_t spsCount{ static_cast<uint8_t>(m_data[pos++] & 0x1F) };
+
+		for (uint8_t i{ 0 }; i < spsCount; i++)
+		{
+			// Get the SPS NALU length
+			uint32_t naluLength{ GetAVCNaluLength(m_data + pos, m_dataSize - pos, sizeof(uint16_t)) };
```
isuru-c-p, line 155: >m_dataSize - pos [](start = 55, length = 16)

The comments from ParseParameterSets apply to this method as well - we'll need a bounds check here and below

FFmpegInterop/NALUSampleProvider.cpp
brbeec: No it's safe. It's just a no-op when naluLength is 0.

FFmpegInterop/HEVCSampleProvider.cpp
```diff
@@ -0,0 +1,120 @@
...
+			{
+				// Reserve estimated space now to minimize reallocations
+				spsPpsNaluLengths.reserve(spsPpsNaluLengths.size() + naluCount);
+			}
+
+			// Scan the NALUs
+			for (uint16_t j{ 0 }; j < naluCount; j++)
+			{
+				// Get the NALU length
+				uint32_t naluLength{ GetAVCNaluLength(m_data + pos, m_dataSize - pos, sizeof(uint16_t)) };
```
isuru-c-p, line 102: >m_dataSize - pos [](start = 56, length = 16)

Need bounds check here and below (in "spsPpsData.insert(spsPpsData.end(), m_data + pos, m_data + pos + naluLength);").

FFmpegInterop/UncompressedAudioSampleProvider.cpp
```diff
@@ -17,161 +17,152 @@
...
-			AV_SAMPLE_FMT_S16,
-			m_pAvCodecCtx->sample_rate,
-			inChannelLayout,
-			m_pAvCodecCtx->sample_fmt,
-			m_pAvCodecCtx->sample_rate,
-			0,
-			NULL);
-
-		if (!m_pSwrCtx)
+		if (m_codecContext->sample_fmt != AV_SAMPLE_FMT_S16)
```
isuru-c-p, line 33: >m_codecContext->sample_fmt != AV_SAMPLE_FMT_S16 [](start = 6, length = 47)

Similar comment to the uncompressed video sample provider - we support more uncompressed audio formats in the MF pipeline.

FFmpegInterop/UncompressedAudioSampleProvider.cpp
```diff
@@ -17,161 +17,152 @@
...
 
-	if (SUCCEEDED(hr))
+	void UncompressedAudioSampleProvider::SetEncodingProperties(_Inout_ const IMediaEncodingProperties& encProp, _In_ bool setFormatUserData)
 	{
-		if (swr_init(m_pSwrCtx) < 0)
-		{
-			hr = E_FAIL;
-		}
+		// We intentionally don't call SampleProvider::SetEncodingProperties() here as
+		// it would set encoding properties with values for the compressed audio type.
```
isuru-c-p, line 57: >values for the compressed audio type. [](start = 43, length = 37)

Shouldn't we be setting the appropriate properties for the uncompressed type here (sample rate, channel count, bitrate and bits per sample)?

FFmpegInterop/UncompressedAudioSampleProvider.cpp
```diff
@@ -17,161 +17,152 @@
...
+				uint8_t* buf{ nullptr };
+				int bufSize{ av_samples_alloc(&buf, nullptr, frame->channels, frame->nb_samples, AV_SAMPLE_FMT_S16, 0) };
+				THROW_HR_IF_FFMPEG_FAILED(bufSize);
 
-MediaStreamSample^ UncompressedAudioSampleProvider::GetNextSample()
-{
-	// Similar to GetNextSample in MediaSampleProvider, 
-	// but we concatenate samples until reaching a minimum duration
-	DebugMessage(L"GetNextSample\n");
+				int resampledSamplesCount{ swr_convert(m_swrContext.get(), &buf, bufSize, const_cast<const uint8_t**>(frame->extended_data), frame->nb_samples) };
```
isuru-c-p, line 114: >bufSize [](start = 69, length = 7)

This existed prior to your refactor, but per the documentation for swr_convert, we should be passing in the number of samples per channel in the output buffer here, instead of the output buffer size in bytes.

FFmpegInterop/UncompressedVideoSampleProvider.cpp
```diff
@@ -18,140 +18,111 @@
...
+			frameRate.Numerator(m_codecContext->framerate.num);
+			frameRate.Denominator(m_codecContext->framerate.den);
 		}
+
+		MediaPropertySet videoProp{ videoEncProp.Properties() };
+		videoProp.Insert(MF_MT_INTERLACE_MODE, PropertyValue::CreateUInt32(MFVideoInterlace_MixedInterlaceOrProgressive));
 	}
 
-	if (SUCCEEDED(hr))
+	tuple<IBuffer, int64_t, int64_t, map<GUID, IInspectable>> UncompressedVideoSampleProvider::GetSampleData()
```
isuru-c-p, line 77: >GetSampleData [](start = 92, length = 13)

Looks like we don't handle format changes (i.e resolution changes). For cases where there frame width / height changes and we're not performing a sws_scale operation (as the pixel format is already NV12), we would send sample data which doesn't conform to the media type exposed on the stream descriptor. And if we are performing a sws_scale operation, we would end up scaling up/down to the old resolution. Neither of these cases are ideal and AFAICT we'd need changes on the OS side in MSS to properly support format changes. For now, we may want to at least prevent the case where we send samples that don't conform to the media type by dynamically initializing the m_swsContext if the frame width/height doesn't match the original width/height (this could probably also be done in a follow up PR since we didn't handle this correctly prior to the refactor either).

FFmpegInterop/UncompressedVideoSampleProvider.cpp
```diff
@@ -18,140 +18,111 @@
...
+			THROW_HR_IF_FFMPEG_FAILED(requiredBufferSize);
+			THROW_HR_IF(MF_E_UNEXPECTED, requiredBufferSize != bufferRef->size);
 
-UncompressedVideoSampleProvider::~UncompressedVideoSampleProvider()
-{
-	if (m_pAvFrame)
-	{
-		av_frame_free(&m_pAvFrame);
-	}
+			THROW_HR_IF_FFMPEG_FAILED(sws_scale(m_swsContext.get(), frame->data, frame->linesize, 0, m_codecContext->height, data, m_lineSizes));
```
isuru-c-p, line 100: >m_codecContext->height [](start = 92, length = 22)

Should use frame->height here (as mentioned in my comment above, the frame width / height can change dynamically).

FFmpegInterop/SampleProvider.cpp
brbeec: Yeah we need to use AVFormatContext::start_time here to avoid the desync issue.

FFmpegInterop/UncompressedAudioSampleProvider.cpp
brbeec: I filed a task.

FFmpegInterop/H264SampleProvider.cpp
brbeec: This is covered by the bounds check in GetAVCNaluLength().

FFmpegInterop/H264SampleProvider.cpp
brbeec: On the first iteration pos is 0 so it's guaranteed to be <= dataSize. On subsequent iterations pos is guaranteed to be <= dataSize due to the bounds check in GetAVCNaluLength().

isuru-c-p: Yea, I missed the bounds check in GetAVCNaluLength().

FFmpegInterop/H264SampleProvider.cpp
isuru-c-p: Resolving as this is also addressed by the bounds check in GetAVCNaluLength

FFmpegInterop/HEVCSampleProvider.cpp
isuru-c-p: Handled by bounds check within GetAVCNaluLength

FFmpegInterop/UncompressedVideoSampleProvider.cpp
brbeec: MSS already supports format changes - on each sample request we check to see if the stream descriptor's encoding properties have been modified. I'll make the necessary changes to handle this case.

PR #274, opened 2020-05-04
Set codec parameters from FLAC stream info in FLACSampleProvider

brbeec:
When FFmpeg is built without the FLAC decoder, some fields (e.g. channels, sample rate, bit rate) may not be set on the stream's codec parameters which can cause playback to fail.

This change updates the FLACSampleProvider to parse the FLAC stream info in the codec private data and set the stream's codec parameters accordingly.

### DIFF
--- a/FFmpegInterop/FLACSampleProvider.cpp
+++ b/FFmpegInterop/FLACSampleProvider.cpp
@@ -18,8 +18,10 @@
 
 #include "pch.h"
 #include "FLACSampleProvider.h"
+#include "BitstreamReader.h"
 
 using namespace winrt::Windows::Foundation;
+using namespace winrt::Windows::Media::MediaProperties;
 using namespace winrt::Windows::Storage::Streams;
 using namespace std;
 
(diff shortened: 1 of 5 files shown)


FFmpegInterop/FLACSampleProvider.cpp
```diff
@@ -28,24 +30,41 @@ namespace winrt::FFmpegInterop::implementation
...
+		bitstreamReader.SkipN(24); // Max frame size
+		m_stream->codecpar->sample_rate = static_cast<int>(bitstreamReader.ReadN(20));
+		m_stream->codecpar->channels = static_cast<int>(bitstreamReader.ReadN(3)) + 1;
+		m_stream->codecpar->bits_per_raw_sample = static_cast<int>(bitstreamReader.ReadN(5)) + 1;
+		// Remaining fields left unparsed as they are unneeded at this time
 	}
 
 	tuple<IBuffer, int64_t, int64_t, vector<pair<GUID, IInspectable>>, vector<pair<GUID, IInspectable>>> FLACSampleProvider::GetSampleData()
 	{
 		if (m_isDiscontinuous)
```
DavidS-msft, line 60: Is this using the MS decoder? Unless the decoder was unloaded it will have an audio glitch if you resend the data.

brbeec: Yes, it's the MS decoder. This is the same behavior as our in-box MKV source and I don't notice any glitches when seeking during playback.

DavidS-msft: We implemented that behavior in the FLAC source and it broke it. So we did the Out of Band media type to pass the information and it fixed that issue.

---
In reply to: [419703605](https://github.com/microsoft/FFmpegInterop/pull/274#discussion_r419703605) [](ancestors = 419703605)

DavidS-msft: However if you are not seeing an issue it might be because you are just sending the stream info and not the entire header. Though if that is the case what happens in Groove when you insert the equalizer effect?

---
In reply to: [419706180](https://github.com/microsoft/FFmpegInterop/pull/274#discussion_r419706180) [](ancestors = 419706180,419703605)

brbeec: The media type attribute used by the FLAC source is private. Not sure why we did that instead of using MF_MT_USER_DATA...

I'm not seeing any issues, but if something comes up we'll need a publicly available solution.

PR #283, opened 2021-03-17
Fix activation of MSS with no stream descriptors

brbeec:
The platform's IMediaStreamSourceFactory implementation will throw an error in its factory methods if a valid IMediaStreamDescriptor is not provided.

This change works around this issue by using IActivationFactory::ActivateInstance() to activate an MSS without a valid IMediaStreamDescriptor.

### DIFF
--- a/FFmpegInterop/FFmpegInteropMSS.h
+++ b/FFmpegInterop/FFmpegInteropMSS.h
@@ -24,7 +24,6 @@
 #include "MediaThumbnailData.h"
 
 using namespace Platform;
-using namespace Windows::Foundation;
 using namespace Windows::Foundation::Collections;
 using namespace Windows::Media::Core;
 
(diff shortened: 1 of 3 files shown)


FFmpegInterop/FFmpegInteropMSS.h
```diff
@@ -24,7 +24,6 @@
 #include "MediaThumbnailData.h"
 
 using namespace Platform;
-using namespace Windows::Foundation;
```
dcarpente, line 27: Why are we removing this, was there some naming collision?

dcarpente: If it's just a stylistic change, I don't mind, I'm approving anyway

dcarpente (approved):

FFmpegInterop/FFmpegInteropMSS.h
brbeec: There were compilation errors from ambiguous symbols after I included wrl.h. In general it's bad practice to write "using namespace" like this in header files. [C++ Core Guidelines SF.7: Don’t write using namespace at global scope in a header file](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf7-dont-write-using-namespace-at-global-scope-in-a-header-file)

PR #299, opened 2023-04-26
MSFT: 43292862, 43115907 - Add script to setup FFmpeg build environment

brbeec:
- Add SetupFFmpegBuildEnvironment.ps1, a script to set up the FFmpeg build environment
- Address [BA2008.EnableControlFlowGuard](https://github.com/microsoft/binskim/blob/main/docs/BinSkimRules.md#rule-BA2008EnableControlFlowGuard) BinSkim errors by building FFmpeg and FFmpegInterop binaries with CFG ([/guard:cf (compiler)](https://learn.microsoft.com/en-us/cpp/build/reference/guard-enable-control-flow-guard?view=msvc-170), [/GUARD:CF (linker)](https://learn.microsoft.com/en-us/cpp/build/reference/guard-enable-guard-checks?view=msvc-170)) enabled
- Address PREfast warnings by fixing SAL annotations

### DIFF
--- a/FFmpegInterop/FFmpegInteropBuffer.cpp
+++ b/FFmpegInterop/FFmpegInteropBuffer.cpp
@@ -55,7 +55,7 @@ namespace winrt::FFmpegInterop::implementation
 		}
 	}
 
-	FFmpegInteropBuffer::FFmpegInteropBuffer(_In_reads_(bufSize) AVBlob_ptr buf, _In_ uint32_t bufSize) :
+	FFmpegInteropBuffer::FFmpegInteropBuffer(_In_ AVBlob_ptr buf, _In_ uint32_t bufSize) :
 		m_length(bufSize),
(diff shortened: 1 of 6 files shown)


FFmpegInterop/FFmpegInteropBuffer.cpp
```diff
@@ -55,7 +55,7 @@ namespace winrt::FFmpegInterop::implementation
 		}
 	}
 
-	FFmpegInteropBuffer::FFmpegInteropBuffer(_In_reads_(bufSize) AVBlob_ptr buf, _In_ uint32_t bufSize) :
```
patnels, line 58: What was wrong with the SAL?

patnels (approved):
LGTM

FFmpegInterop/FFmpegInteropBuffer.cpp
brbeec: SAL doesn't play nice with smart pointers :(

PR #302, opened 2023-10-02
MSFT: 46737360 - Support building FFmpegInterop WACK compliant for Store apps

brbeec:
MFInitMediaTypeFromVideoInfoHeader(), which is used by VFWSampleProvider::SetEncodingProperties() to initialize the stream's VideoEncodingProperties, is not a [WACK supported API](https://learn.microsoft.com/en-us/windows/uwp/debug-test-perf/windows-app-certification-kit-tests).   

## What changed?
This change updates FFmpegInterop to support building WACK compliant for Store apps when WINAPI_FAMILY=WINAPI_FAMILY_PC_APP.

## How was the change tested?
For both WINAPI_FAMILY=WINAPI_FAMILY_PC_APP and WINAPI_FAMILY=WINAPI_FAMILY_DESKTOP_APP, which is the default if WINAPI_FAMILY is not specified, I verified:
- FFmpegInterop builds sucessfully
- Ogg playback is successful in MediaPlayerCPP


xuepingd (approved):
The change looks good.

PR #307, opened 2024-03-05
MSFT: 49309119 - Reduce FFmpegInterop's storage footprint

brbeec:
We want to reduce FFmpegInterop's storage footprint.

## What changed?
- Enable compiler and linker optimizations for favoring small code size
- Remove unnecessary dependencies on avdevice and avfilter

## How was the change tested?
- Ogg playback in MediaPlayerCPP
- Ogg playback in Media Player with WME
- Ogg properties in File Explorer with WME


patnels (approved):
lgtm. smaller layout size, hooray.

PR #308, opened 2024-03-21
MSFT: 47710450 - Add support for building FFmpeg/FFmpegInterop binaries without VCLibs framework package dependencies

brbeec:
We're adding support for building FFmpeg/FFmpegInterop binaries without VCLibs framework package dependencies (e.g. msvcp140_app.dll, vcruntime140_1_app.dll, vcruntime140_app.dll).

Additional reading:
- [Introducing the Universal CRT - C++ Team Blog (microsoft.com)](https://devblogs.microsoft.com/cppblog/introducing-the-universal-crt/)
- [Upgrade your code to the Universal CRT | Microsoft Learn](https://learn.microsoft.com/en-us/cpp/porting/upgrade-your-code-to-the-universal-crt?view=msvc-170&viewFallbackFrom=vs-2019)

## What changed?
- Added `--platform-type` switch to BuildFFmpeg.bat that gets plumbed down to [vcvarsall.bat](https://learn.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-170#vcvarsall-syntax) to set up the build environment. FFmpeg binaries can be built without VCLibs framework package dependencies by **not** specifying `--platform-type store` or `--platform-type uwp`.
- Updated FFmpegInterop project to support building without VCLibs framework package dependencies by setting the UseUCRT property.
- Replaced usage of codecvt, which was deprecated in C++17, with C++/WinRT APIs.
- Fixed x86 build break due to [C4018: signed/unsigned mismatch](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4018?view=msvc-170) warning in FLACSampleProvider::FLACSampleProvider() which became an error after [MSFT: 49309119 - Reduce FFmpegInterop's storage footprint by brbeec · Pull Request #307 · microsoft/FFmpegInterop (github.com)](https://github.com/microsoft/FFmpegInterop/pull/307) set TreatWarningAsError=true.

## How was the change tested?
I validated the following scenarios for build configurations with and without VCLibs framework package dependencies:
- Ogg playback in MediaPlayerCPP
- Ogg playback in Media Player with WME (in-proc and out-of-proc)
- Ogg properties in File Explorer with WME (in-proc and out-of-proc)

I validated the removal of VCLibs framework package dependencies in all of the FFmpeg/FFmpegInterop binaries using depends.exe.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -253,3 +253,11 @@ paket-files/
 # JetBrains Rider
 .idea/
 *.sln.iml
+
+# VS Code
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
+*.code-workspace
--- a/BuildFFmpeg.bat
+++ b/BuildFFmpeg.bat
@@ -148,15 +148,16 @@ if /I %PROCESSOR_ARCHITECTURE% == x86 (
 )
 
 :: Set VSCMD_DEBUG for more verbose debugging output
-:: TODO: Comment out VSCMD_DEBUG
+:: TODO (brbeec):Comment out VSCMD_DEBUG
 set VSCMD_DEBUG=3
 
-:: TODO: Check for UseUCRT flag
-:: TODO: Do we need to call vsdevcmd.bat or can we call vcvarsall.bat without uwp?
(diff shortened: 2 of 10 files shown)


patnels (approved):
lgtm

PR #309, opened 2024-04-09
MSFT: 49847737 - FFmpegInteropBuffer is leaking its underlying buffer

brbeec:
There is a memory leak due to FFmpegInteropBuffer not properly capturing the smart pointer for its underlying buffer in the unique_ptr deleter lambdas, which causes the underlying buffer to not be deallocated when FFmpegInteropBuffer is destroyed.

## What changed?
Updated FFmpegInteropBuffer to properly move capture the smart pointer for its underlying buffer in the unique_ptr deleter lambdas, so that the underlying buffer is deallocated when FFmpegInteropBuffer is destroyed.

## How was the change tested?
I verified under a debugger that the underlying buffer is now being deallocated when FFmpegInteropBuffer is destroyed and memory usage is stable for Ogg playback in Media Player with in-proc WME.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -255,7 +255,7 @@ paket-files/
 *.sln.iml
 
 # VS Code
-.vscode/*
+.vscode/
 !.vscode/settings.json
 !.vscode/tasks.json
 !.vscode/launch.json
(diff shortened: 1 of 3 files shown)


patnels (approved):
lgtm

PR #310, opened 2024-04-11
MSFT: 50055829 - FFmpegInteropByteStreamHandler is closing the byte stream on failure preventing source resolver rollover

brbeec:
FFmpegInteropByteStreamHandler is closing the byte stream on failure which prevents the source resolver from rolling over and attempting other byte stream handlers when the MF_RESOLUTION_CONTENT_DOES_NOT_HAVE_TO_MATCH_EXTENSION_OR_MIME_TYPE flag is set for source resolution.

## What changed?
I added ByteStreamProxy which can be used to wrap a byte stream and prevent it from being closed unless permitted. It forwards all IMFByteStream calls to the underlying byte stream.

FFmpegInteropByteStreamHandler::CreateMediaSource() uses ByteStreamProxy to wrap the byte stream provided to FFmpegInteropByteStreamHandler::BeginCreateObject() and prevent it from being closed unless the MSS is successfully created and initialized, after which the byte stream will be allowed to close when the media source shuts down.

## How was the change tested?
With in-proc WME:
- I verified under a debugger that FFmpegInteropByteStreamHandler no longer attempts to close the byte stream on failure during playback of a corrupted video in Media Player.
- I ran the following TAEF tests:
  - te MF.SourceResolver.unittests.dll /name:SourceResolverTests::TestByteStreamHandlersDontCloseByteStreamOnFailure

### DIFF
--- a/FFmpegInterop/FFmpegInteropByteStreamHandler.cpp
+++ b/FFmpegInterop/FFmpegInteropByteStreamHandler.cpp
@@ -94,7 +94,7 @@ namespace winrt::FFmpegInterop::implementation
 	{
 		auto logger{ FFmpegInteropProvider::CreateMediaSource::Start() };
 
-		// Wrap the byte stream in a proxy to prevent it from being closed if we fail to resolve the media source,
+		// Wrap the byte stream in a proxy to prevent it from being closed if we fail to create and initialize the MSS,
 		// so that the source resolver can rollover and attempt other byte stream handlers.
 		auto byteStreamProxy{ make_self<ByteStreamProxy>(byteStream) };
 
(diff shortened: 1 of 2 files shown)


patnels:
I don't understand why we need this. MSS can be used with non ffmpeg bytestreams, do all bytestreams need this code?

brbeec:
> I don't understand why we need this. MSS can be used with non ffmpeg bytestreams, do all bytestreams need this code?

It's because we use MFCreateStreamOnMFByteStreamEx(); the IRandomAccessStream it returns will close its underlying byte stream when it's destroyed.

Internally MF has similar protections to prevent the byte stream from being closed in various scenarios:
- CMediaEngineByteStreamProxy is used for media engine scenarios to prevent the byte stream from being closed by external callers
- CMediaSourceAppServiceConnection uses an internal interface on the IRandomAccessStream returned by MFCreateStreamOnMFByteStreamEx() to prevent it from being closed unless permitted in out-of-proc media source extension scenarios (e.g. out-of-proc WME)

patnels (approved):

PR #311, opened 2024-04-12
MSFT: 49931784 - Disable FFmpeg logging in release builds

brbeec:
FFmpegInterop.dll registers a custom FFmpeg logging callback, FmpegInteropLogging::Log(), on DLL load. However, this is not safe to do in scenarios that have multiple DLLs using FFmpeg due to the following race condition:
- FFmpegInterop.dll is loaded and registers its custom FFmpeg logging callback, FmpegInteropLogging::Log().
- OtherModule.dll, which also uses FFmpeg, is also loaded in the same process.
- FFmpegInterop.dll is unloaded and [av_log_callback](https://github.com/search?q=repo%3AFFmpeg%2FFFmpeg+av_log_callback&type=code) becomes a dangling pointer.
- OtherModule.dll is still loaded and using FFmpeg. FFmpeg attempts to invoke the now invalid logging callback, resulting in undefined behavior.

Even if the default logging callback was restored on DLL unload, since there is no thread safety in the FFmpeg logging facility, [log_callback in av_vlog()](https://github.com/search?q=repo%3AFFmpeg%2FFFmpeg+log_callback+symbol%3Aav_vlog&type=code) could become a dangling pointer.

## What changed?
Disabled FFmpeg logging in release builds. Debug builds still register the custom logging callback on DLL load, but now also restore the default logging callback on DLL unload (which isn't thread safe and could still result in a crash).

## How was the change tested?
For Ogg playback in Media Player with in-proc WME, I verified under a debugger that:
- For release builds:
  - FFmpeg logging is disabled during DLL load.
- For debug builds:
  - The custom logging callback is registered/unregistered during DLL load/unload.
  - FFmpeg logging is present in ETW traces collected with WPR.

### DIFF
--- a/FFmpegInterop/FFmpegInteropLogging.cpp
+++ b/FFmpegInterop/FFmpegInteropLogging.cpp
@@ -33,6 +33,14 @@ namespace winrt::FFmpegInterop::implementation
 	void FFmpegInteropLogging::Log(_In_ void* avcl, _In_ int level, _In_ const char* fmt, _In_ va_list vl) noexcept
 	try
 	{
+		bool hasLogEventDelegate{ m_logEvent };
+		bool isProviderEnabled{ FFmpegInteropProvider::IsEnabled(WINEVENT_LEVEL_VERBOSE) };
+
+		if (!hasLogEventDelegate && !isProviderEnabled)
+		{
+			return;
+		}
+
 		// Get the required buffer size
 		int printPrefix{ 1 };
 		int lineSize{ av_log_format_line2(avcl, level, fmt, vl, nullptr, 0, &printPrefix) };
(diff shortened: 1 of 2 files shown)


FFmpegInterop/FFmpegInteropLogging.cpp
xuepingd, line 1: Should the following check be '||' instead of '&&'?
if (!hasLogEventDelegate && !isProviderEnabled)

xuepingd (approved):

FFmpegInterop/dllmain.cpp
```diff
@@ -31,11 +31,24 @@ BOOL WINAPI DllMain(_In_ HINSTANCE hInstance, _In_ DWORD dwReason, _In_opt_ LPVO
...
+		av_log_set_callback(nullptr);
+		av_log_set_level(AV_LOG_QUIET);
+#endif // _DEBUG
 	}
 	else if (dwReason == DLL_PROCESS_DETACH)
 	{
+#ifdef _DEBUG
+		// Restore the default FFmpeg logging callback
+		// This is not thread safe! This could leave log_callback in av_vlog() dangling after this DLL is unloaded.
+		av_log_set_callback(av_log_default_callback);
```
patnels, line 49: why isn't it thread safe? can't we fix ffmpeg?

patnels (approved):

FFmpegInterop/dllmain.cpp
brbeec: > why isn't it thread safe?

It's a classic readers-writers problem where the logging callback ([av_log_callback](https://github.com/search?q=repo%3AFFmpeg%2FFFmpeg+av_log_callback&type=code)) is the shared resource:
- Writer: thread setting the logging callback - [av_log_set_callback()](https://github.com/search?q=repo%3AFFmpeg%2FFFmpeg+symbol%3Aav_log_set_callback&type=code)
- Readers: threads invoking the logging callback - [av_vlog()](https://github.com/search?q=repo%3AFFmpeg%2FFFmpeg+symbol%3Aav_vlog&type=code)

You need to synchronize setting the logging callback with threads invoking the callback, otherwise the logging callback could become invalid (i.e. point to memory where our unloaded DLL used to be) while a thread is using it.

> can't we fix ffmpeg?

Yes, we could apply a patch to FFmpeg to address this problem by synchronizing av_log_set_callback() setting the logging callback with av_vlog() accessing the logging callback. This would allow the FFmpeg logging facility to be used safely with our custom logging callback.

Taking a quick peek at the FFmpeg code, it looks like they provide pthread_mutex_* equivalents using [SRWLOCK](https://learn.microsoft.com/en-us/windows/win32/sync/slim-reader-writer--srw--locks) in [w32pthreads.h](https://github.com/search?q=repo%3AFFmpeg%2FFFmpeg+path%3Aw32pthreads.h&type=code), but we would need to add pthread_rwlock_* equivalents using SRWLock.

brbeec: Note that adding synchronization around the logging callback would involve acquiring the writer lock in DllMain() when we call av_log_callback(). This can be perilous as it goes against [DLL best practices](https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-best-practices): "[You should never] synchronize with other threads [from within DllMain]. This can cause a deadlock."

PR #315, opened 2024-05-07
MSFT: 21032904, 50309915, 50626295 - Metadata improvements

brbeec:
- Limited metadata
  - The MSS populates some metadata from the presentation descriptor and stream descriptor(s), but we're not surfacing any embedded metadata (e.g. title, artist, album, etc).
- Thumbnail retrieval race condition
  - The MSS sets the PKEY_ThumbnailStream property for its property handler asynchronously after the MediaStreamSource.Thumbnail property is set. This creates a window of time where a property handler consumer could query for the PKEY_ThumbnailStream property before it's been set and fail to retrieve the thumbnail.
- Hang for files with thumbnails in media source app service scenarios
   - In media source app service scenarios (e.g. out-of-proc WME) the remote stream reference for the thumbnail property could outlive the app service connection. If the remote stream reference is released in the client app after the app service has been suspended, then the remote release will block until the app service is terminated. This could potentially cause the client app to become unresponsive during this time.
- [C6302](https://learn.microsoft.com/en-us/cpp/code-quality/c6302?view=msvc-170) (Format string mismatch) warnings for trace logging
  - We're logging narrow character strings, but using %S which specifies wide character strings.
- Rare crash due to a null MSS reference in FFmpegInteropMSS::OnClosed() for out-of-proc WME on RS4
  - This likely indicates an MSS bug as our MSS reference should only be null if the MediaStreamSource::Closed event handler was already invoked, and the MSS should only fire a single MediaStreamSource::Closed event. However, we've only seen this crash hit on RS4 which is out of support.

## What changed?
- Added support for surfacing embedded metadata (e.g. title, artist, album, etc)
  - The MSS marks some properties read-only which prevents us from setting them.
- Set a thumbnail on the MSS via the PKEY_ThumbnailStream property on the MSS property handler instead of the MediaStreamSource.Thumbnail property to mitigate the race condition with thumbnail retreival
  - Note that the MSS doesn't check its property handler for the PKEY_ThumbnailStream property when a consumer gets the MediaStreamSource.Thumbnail property... This isn't an issue if the MSS consumer is another Windows Media component though, as MF uses the IPropertyHandler interface rather than the IMediaStreamSource interface for retrieving metadata.
- Don't set a thumbnail on the MSS in media source app service scenarios
  -  Media source app services should set the IsMediaSourceAppService property on the FFmpegInteropMSSConfig provided to FFmpegInteropMSS::InitializeFromStream() / FFmpegInteropMSS::InitializeFromUri().
- Updated trace logging to use %hs for narrow character strings
- Use event revokers instead event tokens for unregistering event handlers
  - Event revokers hold a weak reference to the event source (i.e. the MSS) and check if it still exists before revoking the delegate.

## How was the change tested?
- Out-of-proc WME
  - Embedded metadata (e.g. title, artist, album, etc) is surfaced for Ogg files in File Explorer
  - Thumbnails are only generated for Ogg files with video in File Explorer
    - Since out-of-proc WME is a media source app service, it no longer surfaces an thumbnail to the MSS. The MSS will try to generate a thumbnail on-demand for files with a video stream, but files that only have an audio stream will not have a thumbnail.
- In-proc WME scenarios
  - Embedded metadata (e.g. title, artist, album, etc) is surfaced for Ogg files in File Explorer
  - Thumbnails are generated for audio-only Ogg files with an embedded thumbnail in File Explorer


FFmpegInterop/FFmpegInteropLogging.cpp
```diff
@@ -65,7 +65,7 @@ namespace winrt::FFmpegInterop::implementation
 				line[i] = '\0';
 			}
 
-			FFMPEG_INTEROP_TRACE("%S", line.get());
+			FFMPEG_INTEROP_TRACE("%hs", line.get());
```
patnels, line 68: does the to_hstring on line 57 above support char[]?

FFmpegInterop/FFmpegInteropMSS.cpp
```diff
@@ -193,14 +193,19 @@ namespace winrt::FFmpegInterop::implementation
 	{
 		THROW_HR_IF_FFMPEG_FAILED(avformat_find_stream_info(m_formatContext.get(), nullptr));
 
-		int preferredAudioStreamId{ av_find_best_stream(m_formatContext.get(), AVMEDIA_TYPE_AUDIO, -1, -1, nullptr, 0) };
-		int preferredVideoStreamId{ av_find_best_stream(m_formatContext.get(), AVMEDIA_TYPE_VIDEO, -1, -1, nullptr, 0) };
-		bool hasAudio{ false };
-		bool hasVideo{ false };
+		int audioStreamId{ av_find_best_stream(m_formatContext.get(), AVMEDIA_TYPE_AUDIO, -1, -1, nullptr, 0) };
+		int videoStreamId{ av_find_best_stream(m_formatContext.get(), AVMEDIA_TYPE_VIDEO, -1, -1, nullptr, 0) };
+		int thumnbailStreamId{ -1 };
```
patnels, line 198: misspelled 'thumbnail'

FFmpegInterop/FFmpegInteropMSS.cpp
```diff
@@ -335,16 +349,26 @@ namespace winrt::FFmpegInterop::implementation
 		}
 
 		// Populate metadata
-		if (m_formatContext->metadata != nullptr)
+		if (audioStreamId >= 0)
+		{
+			FFMPEG_INTEROP_TRACE("Stream %d: Populating metadata", m_formatContext->streams[audioStreamId]->index);
```
xuepingd, line 354: Maybe "Populating audio metadata"?

FFmpegInterop/Metadata.cpp
```diff
@@ -18,39 +18,140 @@
...
+				unique_prop_variant prop;
+				prop.vt = propKey.type;
+
+				// Convert/parse the property value
+				try
+				{
+					switch (propKey.type)
+					{
+					case VT_LPWSTR:
+						prop.pwszVal = to_cotaskmem_string(tag->value).release();
```
patnels, line 74: does to_cotaskmem_string support char[]?

FFmpegInterop/FFmpegInteropMSS.cpp
```diff
@@ -335,16 +349,26 @@ namespace winrt::FFmpegInterop::implementation
...
+		if (audioStreamId >= 0)
+		{
+			FFMPEG_INTEROP_TRACE("Stream %d: Populating metadata", m_formatContext->streams[audioStreamId]->index);
+			PopulateMetadata(m_mss, m_formatContext->streams[audioStreamId]->metadata);
+		}
+
+		if (videoStreamId >= 0)
 		{
-			PopulateMSSMetadata(m_mss, m_formatContext->metadata);
+			FFMPEG_INTEROP_TRACE("Stream %d: Populating metadata", m_formatContext->streams[videoStreamId]->index);
```
xuepingd, line 360: "Populating video metadata"?

FFmpegInterop/Metadata.cpp
```diff
@@ -18,39 +18,140 @@
...
+		// could potentially cause the client app to become unresponsive during this time. 
+		if (config != nullptr && config.IsMediaSourceAppService())
+		{
+			FFMPEG_INTEROP_TRACE("Stream %d: Not setting thumbnail because this is a media source app service", stream->index);
+			return;
+		}
+
+		FFMPEG_INTEROP_TRACE("Stream %d: Setting thumbnail", stream->index);
+
+		stream->attached_pic.data;
```
patnels, line 121: remove these two lines.

FFmpegInterop/Metadata.cpp
patnels: oh you wrote that helper.

patnels (changes requested):

FFmpegInterop/Metadata.cpp
xuepingd, line 121: Why is stream->attached_pic.data called here, twice?

patnels: why even once? :)

FFmpegInterop/Metadata.h
```diff
@@ -18,8 +18,70 @@
...
+		{
+
+		}
+
+		PROPERTYKEY key;
+		VARTYPE type;
+	};
+
+	inline const std::unordered_map<std::string, PropertyKey> c_metadataPropertiesMap
+	{
```
xuepingd, line 42: Where does this list come from?

FFmpegInterop/Metadata.cpp
xuepingd, line 74: .release() will call CoTaskMemFree()?

FFmpegInterop/Metadata.cpp
brbeec: Unintentional. I must have stopped in the middle of writing something and forgot to clean these lines up.

FFmpegInterop/Metadata.cpp
brbeec: No, release() detaches the PWSTR. reset() would call CoTaskMemFree().

[RAII resource wrappers · Resource Patterns · microsoft/wil Wiki (github.com)](https://github.com/Microsoft/wil/wiki/RAII-resource-wrappers#resource-patterns)
> The interface to all of WIL's RAII / resource management classes is modeled after [std::unique_ptr](http://en.cppreference.com/w/cpp/memory/unique_ptr). You'll find a get() method to access the resource, reset() methods to free or replace the resource and a release() method to detach the raw resource from the management class. See the docs for [std::unique_ptr](http://en.cppreference.com/w/cpp/memory/unique_ptr) for complete reference.

FFmpegInterop/FFmpegInteropMSS.cpp
brbeec: Updated

FFmpegInterop/FFmpegInteropMSS.cpp
brbeec: Updated

FFmpegInterop/FFmpegInteropMSS.cpp
brbeec: Fixed

FFmpegInterop/Metadata.cpp
brbeec: Removed

FFmpegInterop/Metadata.cpp
brbeec: The property handler actually makes a copy of the string when we set the property on it. The original string will be deallocated when our unique_prop_variant goes out of scope, which will result in a call to PropVariantClear() -> CoTaskMemFree().

FFmpegInterop/FFmpegInteropLogging.cpp
brbeec: Yes, there is a templated to_hstring() overload that accepts anything convertible to a std::string_view:

[cppwinrt/strings/base_string.h · microsoft/cppwinrt (github.com)](https://github.com/search?q=repo%3Amicrosoft%2Fcppwinrt%20%22template%20%3Ctypename%20T%2C%20std%3A%3Aenable_if_t%3Cstd%3A%3Ais_convertible_v%3CT%2C%20std%3A%3Astring_view%3E%2C%20int%3E%20%3D%200%3E%22%20%22hstring%20to_hstring(T%20const%26%20value)%22&type=code)
```
template <typename T, std::enable_if_t<std::is_convertible_v<T, std::string_view>, int> = 0>
hstring to_hstring(T const& value)
```

A std::string_view can be constructed from a const char*:

[std::basic_string_view<CharT,Traits>::basic_string_view - cppreference.com](https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view)
> `constexpr basic_string_view( const CharT* s );`
>
> Constructs a view of the null-terminated character string pointed to by s, not including the terminating null character. The length of the view is determined as if by Traits::length(s). The behavior is undefined if [s, s + Traits::length(s)) is not a valid range. After construction, [data()](https://en.cppreference.com/w/cpp/string/basic_string_view/data) is equal to s, and [size()](https://en.cppreference.com/w/cpp/string/basic_string_view/size) is equal to Traits::length(s).

FFmpegInterop/Metadata.cpp
xuepingd: Thanks for clarification.

FFmpegInterop/FFmpegInteropMSS.cpp
```diff
@@ -216,15 +221,15 @@ namespace winrt::FFmpegInterop::implementation
 			case AVMEDIA_TYPE_AUDIO:
 				tie(sampleProvider, streamDescriptor) = StreamFactory::CreateAudioStream(m_formatContext.get(), stream, m_reader, config);
 
-				if (hasAudio || preferredAudioStreamId == i || preferredAudioStreamId < 0)
+				if (i >= audioStreamId)
```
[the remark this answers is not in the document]
brbeec: The logic is same; I just reduced the number of state variables.

**Before:** `hasAudio || preferredAudioStreamId == i || preferredAudioStreamId < 0`
- Add the current audio stream now if we already added an audio stream (`hasAudio`), this is the preferred audio stream (`preferredAudioStreamId == i`), or we don't have a preferred audio stream (`preferredAudioStreamId < 0`).

**After:** `i >= audioStreamId`
- Add the current audio stream now if we already added an audio stream (`i > audioStreamId && audioStreamId >= 0`), this is the preferred audio stream (`i == audioStreamId && audioStreamId >= 0`), or we don't have a preferred audio stream (`audioStreamId < 0`).

FFmpegInterop/Metadata.h
brbeec: The keys come from:
- [metadata_api · FFmpeg/libavformat/avformat.h · FFmpeg/FFmpeg (github.com)](https://github.com/search?q=repo%3AFFmpeg%2FFFmpeg%20metadata_api&type=code)
  - FFmpeg's list of generic tag names
- [Ogg Vorbis Documentation (xiph.org)](https://xiph.org/vorbis/doc/v-comment.html)
  - List of standard tags for Ogg files
- [Tags (matroska.org)](https://www.matroska.org/technical/tagging.html)
  - List of official tags for Matroska files

I referenced internal MF media sources, [Metadata Properties for Media Files - Win32 apps | Microsoft Learn](https://learn.microsoft.com/en-us/windows/win32/medfound/metadata-properties-for-media-files), propkey.h from the Windows SDK, Shell's internal mediaprops.xml, and used my best judgement to come up with the mappings.

xuepingd (approved):

FFmpegInterop/Metadata.cpp
brbeec: Yes, to_cotaskmem_string() accepts anything convertible to std::string_view:
```
template <typename T, std::enable_if_t<std::is_convertible_v<T, std::string_view>, int> = 0>
inline wil::unique_cotaskmem_string to_cotaskmem_string(_In_ const T& str)
```

std::string_view can be constructed from a const char*:

[std::basic_string_view<CharT,Traits>::basic_string_view - cppreference.com](https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view)
> `constexpr basic_string_view( const CharT* s );`
>
> Constructs a view of the null-terminated character string pointed to by s, not including the terminating null character. The length of the view is determined as if by Traits::length(s). The behavior is undefined if [s, s + Traits::length(s)) is not a valid range. After construction, [data()](https://en.cppreference.com/w/cpp/string/basic_string_view/data) is equal to s, and [size()](https://en.cppreference.com/w/cpp/string/basic_string_view/size) is equal to Traits::length(s).

patnels (approved):

PR #322, opened 2024-07-09
MSFT: 50731161 - Enable level 4 warnings

brbeec:
I've been bitten twice by build breaks due to [C4018 (signed/unsigned mismatch)](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4018?view=msvc-170) warnings that repro'd for x86 but not x64 (#308 and #316). It turns out that for the x64 build the warnings being generated were [C4388 (signed/unsigned mismatch)](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c4388?view=msvc-170) which is a level 4, off by default warning.

We want to follow [best practices](https://learn.microsoft.com/en-us/cpp/code-quality/build-reliable-secure-programs?view=msvc-170#23-code-based-or-static-analysis) and enable level 4 warnings.

## What changed?
- Enabled level 4 warnings (`/W4`)
- Enabled [C4388 (signed/unsigned mismatch)](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c4388?view=msvc-170), which is a level 4 warning but off by default (`/w44388`)
- Addressed build breaks due to newly enabled warnings being treated as errors

## How was the change tested?
- Ogg playback in MediaPlayerCPP/MediaPlayerCS
- Ogg playback in Media Player with WME (in-proc and out-of-proc)
- MKV with Vorbis/Theora playback in Media Player with WME (in-proc and out-of-proc)

### DIFF
--- a/FFmpegInterop/Utility.h
+++ b/FFmpegInterop/Utility.h
@@ -56,7 +56,11 @@ namespace winrt::FFmpegInterop::implementation
 {
 	inline std::string tolower(_Inout_ std::string str)
 	{
-		std::transform(str.begin(), str.end(), str.begin(), [](_In_ char c){ return static_cast<char>(std::tolower(c)); });
+		std::transform(str.begin(), str.end(), str.begin(),
+			[](_In_ unsigned char c)
+			{
+				return static_cast<unsigned char>(std::tolower(c));
+			});
 		return str;
 	}
 
--- a/FFmpegInterop/Utility.h
+++ b/FFmpegInterop/Utility.h
@@ -130,7 +130,9 @@ namespace winrt::FFmpegInterop::implementation
 		const int __status = verify_averror(status); \
 		if (__status < 0) \
 		{ \
+			__pragma(warning(disable:4456)) \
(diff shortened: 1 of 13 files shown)


FFmpegInterop/Utility.h
```diff
@@ -56,7 +56,7 @@ namespace winrt::FFmpegInterop::implementation
 {
 	inline std::string tolower(_Inout_ std::string str)
 	{
-		std::transform(str.begin(), str.end(), str.begin(), [](_In_ char c){ return std::tolower(c); });
+		std::transform(str.begin(), str.end(), str.begin(), [](_In_ char c){ return static_cast<char>(std::tolower(c)); });
```
patnels, line 59: nit: looks right to me, but looks different than the sample implementation at https://en.cppreference.com/w/cpp/string/byte/tolower

patnels (approved):
changes look good.

patnels (approved):

PR #326, opened 2024-10-02
MSFT: 53265203 - Remove the workaround in BuildFFmpeg.ps1 for vcvars.bat using the wrong path for OneCore spectre-mitigated libs

brbeec:
I fixed the [bug](https://developercommunity.visualstudio.com/t/vcvarsbat-uses-the-wrong-path-for-OneCo/10664642) with vcvars.bat using the wrong path for OneCore spectre-mitigated libraries, and the fix was released in VS 2022 version 17.11.

## What changed?
I removed the workaround in BuildFFmpeg.ps1 for vcvars.bat using the wrong path for OneCore spectre-mitigated libraries.

## How was the change tested?
I verified BuildFFmpeg.ps1 with `-AppPlatform onecore` worked in the following scenarios:
- A local FFmpeg build with VS 2022 version 17.11.4
- An in-proc WME ADO buddy build with VS 2022 version 17.11.3

### DIFF
--- a/BuildFFmpeg.ps1
+++ b/BuildFFmpeg.ps1
@@ -134,10 +134,6 @@ if (-not ($armasm_flags.Contains('-Q*)')))
     $configure | Out-File -Encoding UTF8 $configurePath
 }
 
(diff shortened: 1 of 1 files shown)


patnels (approved):
lgtm

PR #331, opened 2025-02-27
Add fuzzing flag to FFmpeg build scripts to build FFmpeg with ASAN and SanCov enabled

silvikabra1:
### Why is this change being made?
To create fuzzers for WME (Ogg, Theora, Vorbis), which rely on FFmpeg, we need to be able to build FFmpeg with ASAN and SanCov enabled

### What changed?
To enable fuzzing, we need to add the compiler flags "/fsanitize=address /fsanitize-coverage=inline-8bit-counters /fsanitize-coverage=edge /fsanitize-coverage=trace-cmp /fsanitize-coverage=trace-div" and link to either sancov.lib (dynamic CRT, /MD) or libsancov.lib (static CRT, /MT). The libraries are located in $VCToolsInstallDir\lib\\$arch, which is dependent upon the Visual Studio installation of the building machine. Fuzzing documentation can be found [here](https://eng.ms/docs/cloud-ai-platform/azure-edge-platform-aep/aep-security/epsf-edge-and-platform-security-fundamentals/the-onefuzz-service/onefuzz/faq/notwindows/walkthrough).

I added a fuzzing switch parameter to the PS build script. If fuzzing is enabled, the libs in $VCInstallToolsDir\lib\\$arch get copied over to a different directory. The $VCInstallToolsDir file path has spaces in it, which caused problems while trying to build FFmeg. The new directory with the libraries will be passed in as a parameter to the FFmpeg bash script. The bash script applies the right compiler flags and linker flags to build FFmpeg with ASAN and SanCov enabled if fuzzing is enabled, specifying the library to link based on the CRT used.

I added a fuzzing parameter to the FFmpegInterop.vcxproj file and changed the configuration if that was set to true to include the ASAN compiler flags and link to either sancov.lib or libsancov.lib depending on the RuntimeLibrary. 

ASAN doesn't compile with EditAndContinue (/ZI), so I had to change the DebugInformationFormat to ProgramDatabase (/Zi) for Debug configurations.
There was a level 4 warning in VFWSampleProvider.cpp so I added a static cast to fix that.

### How was the change tested?
Built FFmpeg with x64 and fuzzing flag set to true
Built FFmpegInterop with the following command: MSBuild FFmpegInterop.sln -t:Clean;Build /p:Configuration=Debug /p:Platform=x64 /p:Fuzzing=true

### DIFF
--- a/BuildFFmpeg.ps1
+++ b/BuildFFmpeg.ps1
@@ -175,7 +175,7 @@ foreach ($arch in $Architectures)
             exit 1
         }
         Write-Host "Adding $fuzzingLibPath to the LIB environment variable"
-        $env:LIB="$env:LIB;$fuzzingLibPath"
+        $env:LIB = "$env:LIB;$fuzzingLibPath"
         $opts += '--fuzzing'
     }
 
--- a/BuildFFmpeg.ps1
+++ b/BuildFFmpeg.ps1
@@ -66,7 +66,7 @@ param(
 
     [string]$Settings = '',
 
-    [bool]$Fuzzing = $false
+    [switch]$Fuzzing
 )
 
 # Validate FFmpeg submodule
@@ -176,7 +176,7 @@ foreach ($arch in $Architectures)
         }
 
         # Copy the contents of $VCInstallToolsDir\lib\arch\* to $PSScriptRoot\libs 
-        Write-Host "Copying fuzzing libs from $env:VCToolsInstallDir to $fuzzingLibsDir"
+        Write-Host "Copying fuzzing libs from $env:VCToolsInstallDir\lib\$arch to $fuzzingLibsDir"
         Copy-Item -Path "$env:VCToolsInstallDir\lib\$arch\*" -Destination $fuzzingLibsDir -Recurse -Force
         
         # FFmpeg build needs forward slashes in paths
--- a/BuildFFmpeg.ps1
+++ b/BuildFFmpeg.ps1
@@ -165,23 +165,18 @@ foreach ($arch in $Architectures)
         $opts += '--settings', $Settings
     }
 
-    # Fuzzing requires libraries in the $VCToolsInstallDir to be lnked
+    # Fuzzing requires libraries in the $VCToolsInstallDir to be linked
     if ($Fuzzing)
     {
-        # Create new directory since $VCToolsInstallDir has spaces
-        $fuzzingLibsDir = Join-Path $PSScriptRoot 'libs'
-        if (-not (Test-Path $fuzzingLibsDir))
+        $fuzzingLibPath = Join-Path -Path $env:VCToolsInstallDir -ChildPath "lib\$arch"
+        if (-not (Test-Path $fuzzingLibPath))
         {
-            New-Item -ItemType Directory -Path $fuzzingLibsDir
+            Write-Error "ERROR: $fuzzingLibPath does not exist. Ensure the Visual Studio installation is correct."
+            exit 1
         }
-
-        # Copy the contents of $VCInstallToolsDir\lib\arch\* to $PSScriptRoot\libs 
-        Write-Host "Copying fuzzing libs from $env:VCToolsInstallDir\lib\$arch to $fuzzingLibsDir"
-        Copy-Item -Path "$env:VCToolsInstallDir\lib\$arch\*" -Destination $fuzzingLibsDir -Recurse -Force
-        
-        # FFmpeg build needs forward slashes in paths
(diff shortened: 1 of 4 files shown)


silvikabra1:
@microsoft-github-policy-service agree company="Microsoft"

BuildFFmpeg.ps1
```diff
@@ -171,6 +195,13 @@ foreach ($arch in $Architectures)
         Set-Item "env:$($envVar.Name)" $envVar.Value
     }
 
+    # Delete the contents of the fuzzing libs directory and the folder itself
+    if (Test-Path $fuzzingLibsDir)
+    {
+        Remove-Item -Path "$fuzzingLibsDir\*" -Recurse -Force
+        Remove-Item -Path $fuzzingLibsDir
```
brbeec, line 202: If we recursed on the directory removal, would we still need to remove the contents first? (i.e. can the two lines above be combined into one)

BuildFFmpeg.ps1
```diff
@@ -160,6 +165,25 @@ foreach ($arch in $Architectures)
         $opts += '--settings', $Settings
     }
 
+    # Fuzzing requires libraries in the $VCToolsInstallDir to be lnked
```
brbeec, line 168: typo: "lnked"

FFmpegConfig.sh
```diff
@@ -129,14 +135,34 @@ else
...
+    fi
+
+    fuzz_settings="
+        --extra-cflags=\"-fsanitize=address -fsanitize-coverage=inline-8bit-counters -fsanitize-coverage=edge -fsanitize-coverage=trace-cmp -fsanitize-coverage=trace-div\" \
+        --extra-ldflags=\"-LIBPATH:$fuzzing_libs\" \
+        "
+
+    # Add sancov.lib or libsancov.lib based on CRT
+    if [[ $crt == "dynamic" ]]; then
+        fuzz_crt_settings="--extra-ldflags=\"-DEFAULTLIB:sancov.lib\""
```
brbeec, line 152: nit: suggest just appending to fuzz_settings rather than setting an additional variable

brbeec:
We should also update FFmpegInterop.vcxproj in this change to support building FFmpegInterop.dll with ASAN and SanCov enabled

BuildFFmpeg.ps1
```diff
@@ -160,6 +165,25 @@ foreach ($arch in $Architectures)
         $opts += '--settings', $Settings
     }
 
+    # Fuzzing requires libraries in the $VCToolsInstallDir to be lnked
+    if ($Fuzzing)
+    {
+        # Create new directory since $VCToolsInstallDir has spaces
+        $fuzzingLibsDir = Join-Path $PSScriptRoot 'libs'
```
brbeec, line 172: You're assuming that $PSScriptRoot doesn't have any spaces, which may not always be a valid assumption

BuildFFmpeg.ps1
silvikabra1: Fixed -- changed solution and added the original directory with the necessary libs to the environment variable $env:LIB, which works.

BuildFFmpeg.ps1
silvikabra1: Fixed

FFmpegConfig.sh
silvikabra1: Fixed

BuildFFmpeg.ps1
silvikabra1: True. Changed solution to add $VCToolsInstallDir to the $env:LIB environment variable, which works

BuildFFmpeg.ps1
```diff
@@ -160,6 +165,20 @@ foreach ($arch in $Architectures)
...
+    if ($Fuzzing)
+    {
+        $fuzzingLibPath = Join-Path -Path $env:VCToolsInstallDir -ChildPath "lib\$arch"
+        if (-not (Test-Path $fuzzingLibPath))
+        {
+            Write-Error "ERROR: $fuzzingLibPath does not exist. Ensure the Visual Studio installation is correct."
+            exit 1
+        }
+        Write-Host "Adding $fuzzingLibPath to the LIB environment variable"
+        $env:LIB="$env:LIB;$fuzzingLibPath"
```
brbeec, line 178: nit: add spacing around assignment operator

brbeec (approved):

BuildFFmpeg.ps1
silvikabra1: fixed

brbeec (approved):

PR #333, opened 2025-04-17
MSFT: 54032803 - Add support for enabling PREfast and EO-compliant compiler warnings for FFmpeg builds

brbeec:
We're adding support for enabling PREfast static analysis and EO-compliant compiler warnings for FFmpeg builds.

## What changed?
- Added new optional parameters to BuildFFmpeg.ps1 and FFmpegConfig.sh:
  - Prefast - Specifies a ruleset to use for PREfast static analysis
    - See [/analyze (Code analysis) | Microsoft Learn](https://learn.microsoft.com/en-us/cpp/build/reference/analyze-code-analysis) for more info about PREfast static analysis.
  - Patches - Specifies one or more patches or directories containing patches to apply to FFmpeg before building
  - CompilerRsp - Specifies a compiler response files (.rsp) to pass to the FFmpeg build
    - See [@ (Specify a Compiler Response File) | Microsoft Learn](https://learn.microsoft.com/en-us/cpp/build/reference/at-specify-a-compiler-response-file) for more info about compiler response files.
  - SarifLogs - Specifies whether to enable SARIF output diagnostics for MSVC
    - See [Structured SARIF Diagnostics | Microsoft Learn](https://learn.microsoft.com/en-us/cpp/build/reference/sarif-output) for more info about SARIF diagnostics.
- Created 000_armasm_flags.patch to modify [armasm_flags()](https://github.com/search?q=repo%3AFFmpeg%2FFFmpeg%20armasm_flags&type=code) in FFmpeg's configure script to filter out additional MSVC compiler options and their arguments from being passed to gas-preprocessor.pl in ARM/ARM64 builds to avoid A2029 (unknown command-line argument) errors

## How was the change tested?
I built FFmpeg for x86, x64, and arm64 using BuildFFmpeg.ps1 with and without the new optional parameters.

### DIFF
--- a/FFmpegConfig.sh
+++ b/FFmpegConfig.sh
@@ -175,7 +175,7 @@ if [[ $fuzzing ]]; then
 
     fuzz_settings="\
         --extra-cflags=\"-fsanitize=address -fsanitize-coverage=inline-8bit-counters -fsanitize-coverage=edge \
-        -fsanitize-coverage=trace-cmp -fsanitize-coverage=trace-div\""
+        -fsanitize-coverage=trace-cmp -fsanitize-coverage=trace-div\" "
 
     # Add sancov.lib or libsancov.lib based on CRT
     if [[ $crt == "dynamic" ]]; then
--- a/FFmpegConfig.sh
+++ b/FFmpegConfig.sh
@@ -2,7 +2,7 @@
 dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 
 # Parse the options
-options=$(getopt -o "" --long arch:,app-platform:,settings:,crt:,prefast:,sarif-logs,fuzzing -n "$0" -- "$@")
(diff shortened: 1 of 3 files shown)


FFmpegConfig.sh
```diff
@@ -96,59 +116,65 @@ common_settings=" \
...
         --extra-cflags=\"-D__ARM_PCS_VFP\" \
-        --prefix=../../Build/$arch \
+        --prefix=$dir/ffmpeg/Build/$arch \
         "
 else
     echo "ERROR: $arch is not a valid architecture" 1>&2
     exit 1
 fi
 
+if [[ $sarif_logs ]]; then
```
silvikabra1, line 157: nit: Could get rid of the sarif_logs variable and move the variable and definition of sarif_logs_setting to line 87 where we parse the parameters passed in the command line, just like how prefast_settings and compiler_rsp_settings get set. I didn't do this for fuzz_settings because it relied upon $arch and $crt

silvikabra1 (approved):

FFmpegConfig.sh
brbeec: sarif_logs_settings also relies on $arch which is why I set it outside of the options parsing loop

